PR #293, opened 2021-01-15
Merge to main

jonseitz:
As noted for the corresponding "merge to main" PR on seas-computing/mark-one#85, this is just wrapping up a bunch of code that's already been reviewed before merging into develop, so the review request is more about confirming the merge into main than it is about reviewing the code. 

Travis started failing again when I made the repo private again, so I'm going to leave it public for now and we'll see how that goes.


rmainseas (approved):
This is good to go.

PR #294, opened 2021-01-19
Update Metadata GET to include Catalog Prefixes

### LINKED ISSUE #292: Add catalog prefixes to the metadata.

natalynyu:
This PR updates the metadata `GET` request to include the current catalog prefixes that exist in the database. In order to see the metadata output, you can run docker and go to `http://localhost:3001/course-planner/api/metadata`. 

image

## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #292


src/server/course/course.service.ts
```diff
@@ -30,4 +30,24 @@ export class CourseService {
...
+
+  /**
+   * Resolve an array containing all catalog prefixes that currently exist in the
+   * database, as strings
+   */
+  public async getCatalogPrefixList(): Promise<string[]> {
+    return this.courseRepository
+      .createQueryBuilder('c')
+      .select('c.prefix', 'prefix')
+      .distinct(true)
```
jonseitz, line 42: One issue here is that in the four-year plan interface where we'll be using this data, the courses in the list are only the SEAS/EPS courses, as defined in the [MultiYearPlanView entity](https://github.com/seas-computing/course-planner/blob/2cc517d570680cb5cf8f33610d38c545e3a0fcf9/src/server/courseInstance/MultiYearPlanView.entity.ts#L21). If this part of the metadata is only going to be used in the four-year plan and the schedule (which also only shows SEAS courses) it probably makes sense to limit this query to only courses where `"isSEAS" <> IS_SEAS.N`.

src/common/__tests__/data/courses.ts
```diff
@@ -117,3 +117,22 @@ export const updateCourseExample: UpdateCourseDTO = {
...
+  { prefix: 'BE' },
+  { prefix: 'CS' },
+  { prefix: 'EE' },
+  { prefix: 'ESE' },
+  { prefix: 'General' },
+  { prefix: 'Mat & ME' },
+  { prefix: 'MDE' },
+  { prefix: 'MSMBA' },
+  { prefix: 'SEM' },
+];
```
jonseitz, line 138: A couple of these (e.g. `'EE'`, `'Mat & ME'`) aren't actually course prefixes, but rather Areas.  It's a minor point, but as always we want our test data to mirror reality as much as possible.

src/common/__tests__/data/metadata.ts
```diff
@@ -20,6 +20,11 @@ const semesters = [
   'Spring 2022',
 ];
 
+/**
+ * Example of a list of catalog prefixes
+ */
+const catalogPrefixes = ['AC', 'AM', 'AP', 'BE', 'CS', 'EE', 'ESE', 'General', 'Mat & ME', 'MDE', 'MSMBA', 'SEM'];
```
jonseitz, line 26: See above re: prefixes vs. areas.

jonseitz (changes requested):
Two minor notes about the test data below, along with a note about limiting the list of prefixes to only those within SEAS. I considered suggesting that we could create two lists of prefixes, one with all of theme and the other with only the SEAS ones, but I don't think we have a clear use case for the "all prefixes" list so better to stick with just what we need now. Happy to discuss more if you feel differently

src/common/__tests__/data/courses.ts
natalynyu: Ah good point! I should have checked the actual app. I updated the test data in 14cd59c so that it matches the current app's catalog prefixes in the "4 Year Plan" tab.

src/common/__tests__/data/metadata.ts
natalynyu: This was also updated in 14cd59c.

src/server/course/course.service.ts
natalynyu: Oh, I see, ok. I updated the course service in 3636a96 along with two course service test setups to account for the new `where` clause.

jonseitz (approved):
This is all set!

PR #298, opened 2021-01-27
Multi-Year Plan Filters and update MYP client to handle insufficient semester rows

### LINKED ISSUE #281: 4-year plan table filters
Implements the dropdown and free text filter fields that have been added to the faculty admin table in #260

Note: Course catalog prefix should be a dropdown rather than a free text as indicated in the mockup.

natalynyu:
This PR adds the multi-year plan filters. The number of instructor filters that are created is based off of the number of semesters that are returned in the multi year plan list. The catalog prefix dropdown is populated by the catalog prefixes in the metadata.

Also, the multi-year plan GET controller action was updated so that regardless of the number of semesters rows that exist in the database, we will receive the number of semesters that were requested. For instance, if there are 6 semester rows and we request 8, the controller now creates the empty semester objects to supplement. Previously, the controller would return as many semesters as possible if it didn't have as many as were requested, and we had assumed that the database would always have enough semester rows. Instead, as an improvement, this update to the controller guarantees we will get the number of semesters requested. 

EDIT -- In the interest of keeping the server and client logic more separate as noted/recommended by J Seitz in the PR comments, we decided to let the controller return what it gets from the server and have the client handle cases in which the controller returns fewer semesters than requested.

## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #281


jonseitz (changes requested):
Seeing this all laid out in code, I think I may have steered you down the wrong path by suggested that the server should return empty values when there aren't semesters. I think you had a good solution to just add placeholder objects for the missing semester values, but I'm a little worried about long-term maintainability. 

So I suggested in a couple notes below that we should have a single constant for the number of years or semesters, and use that separately in both the client and server code and keep the concerns of each side separate: The server can return whatever data it has, and the client will render as many columns as it wants from that. What do you think about that?

src/client/components/pages/MultiYearPlan.tsx
```diff
@@ -79,6 +143,36 @@ const MultiYearPlan: FunctionComponent = (): ReactElement => {
...
 
+  /**
+   * Creates the instructor filters using the multi year plan list
+   */
+  const instructorFilters = (myp: MultiYearPlanResponseDTO[]):
+  TableHeadingCell[] => (
+    myp.length > 0
+      ? myp[0].semesters
+        .map((semester: MultiYearPlanSemester, index: number):
+        TableHeadingCell => (
```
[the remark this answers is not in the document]
natalynyu: That makes sense! I added `NUM_YEARS` and `NUM_SEMESTERS` to the `common/constants` so that these values could be used to calculate the right number of columns.

src/server/courseInstance/courseInstance.controller.ts
```diff
@@ -97,9 +104,58 @@ export class CourseInstanceController {
...
+    const missingSemesters: MultiYearPlanSemester[] = Array.from(
+      new Array(missingNumSemesters)
+    ).map((_, index) => (
+      createMissingSemester(index)
+    ));
+    if (missingNumSemesters > 0) {
+      multiYearPlans = multiYearPlans.map((plan) => ({
+        ...plan,
+        semesters: [...plan.semesters, ...missingSemesters],
+      }));
```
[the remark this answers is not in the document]
natalynyu: @jonseitz Thanks for this suggestion! I think this makes a lot of sense. I reverted the controller logic back to return what is available in the db in commit 830ff75 along with updating the client to use the constants in the `common/constants` directory to calculate the right number of columns/filters.

I noticed the build is not running in this PR - does having a merge conflict prevent the build from running? Should I try to resolve the merge conflict now before re-review?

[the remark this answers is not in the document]
jonseitz: Ah, so the builds aren't running because the workflow files don't exist in this branch yet. You'd need to merge develop back into here before those will run.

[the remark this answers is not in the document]
natalynyu: D'oh! Thanks Jonathan.

src/common/utils/multiYearPlanHelperFunctions.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Gets the list of semesters for the multi-year plan table starting with the
+ * given year and specified number of semesters.
+ * This function was separated from getSemestersFromYear to allow easier unit
+ * testing.
+ */
+export const calculateSemesters = (
+  startingAcademicYear: number,
+  numSemesters: number
+):
+SemesterInfo[] => range(numSemesters)
```
jonseitz, line 21: See note above about using `Array.from()` for this

src/common/utils/__tests__/multiYearPlanHelperFunctions.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    it('returns the requested number of semesters', function () {
+      const actual = calculateSemesters(
+        metadata.currentAcademicYear, NUM_SEMESTERS
+      );
+      strictEqual(actual.length, NUM_SEMESTERS);
+    });
+    it('returns the correct terms', function () {
+      const years = Array(NUM_YEARS);
+      const terms = [TERM.FALL, TERM.SPRING];
+      const expectedTerms = flatten(years.fill(terms));
```
jonseitz, line 24: Along the lines of my previous note about `lodash`, there are a couple of workarounds for [flattening a single-level array[(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#reduce_and_concat), the easiest being:

```ts
const expectedTerms = [].concat(...years.fill(terms));
```

We're also going to need to upgrade to Node v14 soon anyway, at which point we can use `Array.flat()` and `Array.flatMap()` out of the box

src/common/utils/__tests__/multiYearPlanHelperFunctions.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      const actualTerms = multiYearPlan.calculateSemesters(
+        metadata.currentAcademicYear,
+        NUM_SEMESTERS
+      ).map((semester) => semester.term);
+      deepStrictEqual(actualTerms, expectedTerms);
+    });
+    it('returns the expected academic years', function () {
+      const expectedSemesterAcademicYears = [2021, 2021, 2022, 2022,
+        2023, 2023, 2024, 2024];
```
jonseitz, line 33: Out style is to split long arrays like this with one element on each line.

src/common/utils/__tests__/multiYearPlanHelperFunctions.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      const actualAcademicYears = multiYearPlan.calculateSemesters(
+        metadata.currentAcademicYear,
+        NUM_SEMESTERS
```
jonseitz, line 36: Instead of using the `NUM_SEMESTERS` constant here, I'd suggest substituting a literal `8` (or any other even number). That way if we ever do need to change the constant to make this a five-year plan, we won't break all these tests in the process.

src/client/components/pages/utils/filterByInstructorValues.ts
```diff
@@ -0,0 +1,32 @@
...
+    .forEach(([index, instructorFilter]) => {
+      const instructorFilterLower = instructorFilter.toLowerCase();
+      // Converting index to a number to appease compiler, which thinks index
+      // is a string, but we know that index will be a number.
+      const indexNum = parseInt(index, 10);
+      // Filter by the instructors
+      plans = plans
+        .filter((plan) => plan
+          .semesters[indexNum]
+          .instance.faculty.some((instructor) => instructor.displayName
```
jonseitz, line 25: This line is throwing an error when I try to search within a column that does't have any data associated with it (i.e., there's no corresponding semester in the database).

src/client/components/pages/MultiYearPlan.tsx
```diff
@@ -85,19 +181,22 @@ const MultiYearPlan: FunctionComponent = (): ReactElement => {
...
-              </TableCellListItem>
-            ))
-            : null}
-        </TableCellList>
-      </TableCell>
-    ))
+    range(NUM_SEMESTERS).map((index: number) => {
+      const semester = course.semesters[index];
+      return (
+        <TableCell verticalAlignment={VALIGN.TOP} key={index}>
```
jonseitz, line 187: It's generally not advised to use array indexes as keys -- what about using the faculty `id` field with the `term` and `academicYear` appended to it?

jonseitz (changes requested):
I noted a couple places below where the added lodash functions aren't totally necessary. There's also a bug when you try to search on a column that doesn't have any corresponding data in the database, where we might need to add some additional checking. 

Overall this is looking really good though!

src/client/components/pages/MultiYearPlan.tsx
```diff
@@ -85,19 +181,22 @@ const MultiYearPlan: FunctionComponent = (): ReactElement => {
...
-            ? semester.instance.faculty.map((faculty): TableCellListItem => (
-              <TableCellListItem key={faculty.id}>
-                {faculty.displayName}
-              </TableCellListItem>
-            ))
-            : null}
-        </TableCellList>
-      </TableCell>
-    ))
+    range(NUM_SEMESTERS).map((index: number) => {
```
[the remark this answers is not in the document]
natalynyu: Oh nice, ok sounds good! The uses of `lodash.range` were replaced in e2dea8e.

src/common/utils/multiYearPlanHelperFunctions.ts
natalynyu: This was also replaced in e2dea8e.

src/common/utils/__tests__/multiYearPlanHelperFunctions.test.ts
natalynyu: Ah, didn't know about this. Thanks! This was updated in e2dea8e.

src/common/utils/__tests__/multiYearPlanHelperFunctions.test.ts
natalynyu: Oops! Ok the formatting was fixed in 806778a.

src/common/utils/__tests__/multiYearPlanHelperFunctions.test.ts
natalynyu: Good idea! I tried making each test more insular but missed this one. This was fixed by setting the number of semesters to `8` in 806778a.

src/client/components/pages/MultiYearPlan.tsx
natalynyu: I thought it was ok to use the index as the key since the semester columns would be static? I realized though that I could use the `getSemestersFromYear` function instead of constructing the semesters in the original implementation, so I am now using the semester key, which is `academicYear-term.` (commit 75b6b85).

src/client/components/pages/utils/filterByInstructorValues.ts
natalynyu: Thanks for catching this! I believe that the check for the truthiness of `plan.semesters[indexNum]` that I added in commit b4ff693 will fix this.

jonseitz: Yep, that did the trick!

jonseitz (approved):
This looks good! I think it's ready to be merged.

PR #319, opened 2021-03-16
Update Meeting Modal Formatting

### LINKED ISSUE #305: Courses table: Room booking modal furniture
When the edit icon in the "Meetings" column in the course instance table is clicked, open a modal using the Mark One modal component.

Includes:

* [x] Modal header, with title in the format of "Rooms for {catalog prefix and number, e.g. CS 50} - {course title, e.g. }"
* [x] Modal footer, including Cancel button
  * Clicking cancel button closes the modal
* [x] Modal body, including CSS Grid for interior components

natalynyu:
This PR updates the formatting of the meeting modal to have two sections, Meeting Scheduler and Room Availability, instead of three (originally there was a Faculty Notes section). I used flexbox in order to get the two divs to scroll independently instead of having the scrollbar on the entire modal body - I struggled to find a way to accomplish the independent scrolling of sections with CSS grid but am welcome to suggestions if CSS grid is the preference. Also, each section's headings are sticky so that when the user scrolls, the header is still visible.

Below are some screenshots of the two sections filled with sample content:

Screen Shot 2021-03-16 at 1 31 33 PM
Screen Shot 2021-03-16 at 2 01 09 PM
Screen Shot 2021-03-16 at 2 01 42 PM

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #305

### DIFF
--- a/src/client/components/pages/Courses/MeetingModal.tsx
+++ b/src/client/components/pages/Courses/MeetingModal.tsx
@@ -37,37 +37,61 @@ interface MeetingModalProps {
  * Utility component to style content within meeting modal body
  */
 const MeetingModalBodyGrid = styled.div`
-  display: grid;
-  grid-template-columns: repeat(2, 1fr);
-  grid-template-rows: repeat(2, 1fr);
-  grid-template-areas:
-  "meetingScheduler roomAvailability"
-  "facultyNotes roomAvailability";
+  height: 75vh;
+  width: 75vw;
+  display: flex;
+  flex-direction: row;
   column-gap: ${fromTheme('ws', 'xsmall')};
 `;
 
 /**
- * Placeholder for the meeting scheduler form of the meeting modal for
- * display purposes
+ * Contains the meeting scheduler section of the meeting modal
  */
 const MeetingScheduler = styled.div`
-  grid-area: meetingScheduler;
+ display: flex;
+ flex-direction: column;
+ flex: 1;
 `;
 
 /**
- * Placeholder for the faculty notes section of the meeting modal for
- * display purposes
+ * The header of the meeting scheduler section of the Meeting Modal
  */
-const FacultyNotes = styled.div`
-  grid-area: facultyNotes;
+const MeetingSchedulerHeader = styled.h3`
+ flex: 0;
+`;
+
+/**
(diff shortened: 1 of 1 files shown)


src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -37,37 +37,61 @@ interface MeetingModalProps {
...
  */
 const MeetingModalBodyGrid = styled.div`
-  display: grid;
-  grid-template-columns: repeat(2, 1fr);
-  grid-template-rows: repeat(2, 1fr);
-  grid-template-areas:
-  "meetingScheduler roomAvailability"
-  "facultyNotes roomAvailability";
+  height: 75vh;
+  width: 75vw;
```
jonseitz, line 41: I think these should be `max-height` and `max-width` instead of absolute values, so we don't end up with a modal that's significantly larger than its content.

src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -37,37 +37,61 @@ interface MeetingModalProps {
...
+  display: flex;
+  flex-direction: row;
   column-gap: ${fromTheme('ws', 'xsmall')};
```
jonseitz, line 44: This can be removed, since we're not using `grid` anymore

jonseitz (changes requested):
Two small notes on the css below. 

I also think we should have a larger gap between the two columns -- in some of those screenshots the text is almost touching the adjoining column. I think we could do that with `justify-content: space-between;` on the parent, along with a change to the `flex: 1` rule on the columns to set `flex-basis` to something something smaller (e.g. `flex: 48%;`)

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Good point! I replaced the `height` and `width` with `max-height` and `max-width.`

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Thanks for pointing this out. This was removed in c001d11.

jonseitz (approved):
This is all set!

PR #326, opened 2021-04-06
Codecov Updates

jonseitz:
The changes in this PR came out of the last developer meeting (03/31) between @jonseitz, @natalynyu, and @rmainseas:

1. **Coverage checks will no longer be required for a PR to pass.** This was leading to a lot of situations where we were adding less-than-totally-necessarily tests just to get the coverage stats up to a level that would pass our CI check, and that's not how we want to be approaching our tests. There should still be informational reports shown in the PR status, but the checks themselves should always pass.
2. **The test run will no longer fail if coverage is below a certain threshold.** This was a less common source of issues because our thresholds were fairly low (80%), but still worth turning off in light of 1 above.
3. **The coverage report will no longer print out automatically at the end of test runs.** This a developer experience improvement, as the project has grown so large that the coverage report (inessential information) was pushing the test results (essential information) out of the terminal.
4. **New scripts have been added to show the coverage report in text/html format.**
  a. `npm run cov:web` should open the html report in the browser.
  b. `npm run cov:text` will print the text report in the terminal.

On point 4, it's important to note that **the coverage will only be accurate as of the last test run; you will need to re-run the test suite to update the report.**

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High


natalynyu (approved):
This looks great! Approved.

PR #328, opened 2021-04-13
Feature/210 non class event views

jonseitz:
**Marking as a breaking change because it requires a new migration to be run**

This PR splits out the `NonClassEventView` and `NonClassParentView` from @rmainseas's PR #232, as well as the relationship added there between the `NonClassEventView` and the `MeetingListingView`. Moving the code out did require some rebasing to remove some unrelated changes, and I've added references to the commit messages to provide a sort of audit trail of what was consolidated and/or moved.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality
      to change)

## Checklist:

- [x] I have run `eslint` on the code
- [x] I have added JSDoc for all of my code (where applicable)
- [ ] I have added tests to cover my changes.

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

unblocks #304

PR #329, opened 2021-04-15
Bugfix/view relationships

jonseitz:
This changes the definition for the `ScheduleEntryView` so that it no longer relies on a JOIN with the `RoomListingView` and the `MeetingListingView`. Those relationships were causing the migration included in #328 to fail, because TypeORM handles changes to views by dropping and recreating them, rather than altering them in place.

Note that I did manually edit the timestamp of the new migration to force it to run before `1618340375982-NonClassEventViews.ts`. That seemed like the least invasive way to resolve the dependency issue. If you've somehow already run that migration, you'll need to `migration:revert` it before running this new one, or simply "nuke and pave" the postgres container and run all the migrations on an empty database.  

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High

### DIFF
--- a/src/server/migrations/1618340375981-RemoveViewDependencies.ts
+++ b/src/server/migrations/1618340375981-RemoveViewDependencies.ts
@@ -1,0 +1,33 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * This migration replaces the ScheduleEntryView created in
+ * 1603825601596-ScheduleViews.ts to remove the relation between the
+ * ScheduleEntryView and the MeetingListingView and RoomListingView, which was
+ * creating a dependency in the view definition and making it impossible to
+ * update the MeetingListingView in 1618340375982-NonClassEventViews.ts.
+ * Becuase that dependency breaks the migration:run process, the timestamp on
+ * this migration has been modified so that it will run BEFORE the
+ * NonClassEventViews migration.
+ *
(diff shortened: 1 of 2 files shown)


src/server/migrations/1618340375981-RemoveViewDependencies.ts
```diff
@@ -0,0 +1,33 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * This migration replaces the ScheduleEntryView created in
+ * 1603825601596-ScheduleViews.ts to remove the relation between the
```
rmainseas, line 5: I would add square brackets round this, so that it can be clicked through in the documentation. That's also true for the other views, but that's probably less less important than having a reference to the other migration.

```suggestion
 * [[1603825601596-ScheduleViews.ts]] to remove the relation between the
```

rmainseas (approved):
This looks great to me. There one minor thing regarding documentation, but it's really not a big deal - so I'm approving this regardless.

src/server/migrations/1618340375981-RemoveViewDependencies.ts
jonseitz: Not a bad idea, but the migrations are actually [ignored by our typedoc config](https://github.com/seas-computing/course-planner/blob/main/typedoc.json#L6) so they won't show up in the generated docs anyway.

PR #335, opened 2021-05-03
Fix Display of Faculty Notes when 'notes' is null - Meeting Modal

### LINKED ISSUE #331: Meeting Modal - Faculty Notes bugfix
When the notes of a faculty member is null, it should display "No Notes" under the faculty name in the meeting modal. This is currently not the case, because the code is checking only checking whether the 'notes' value is an empty string.

natalynyu:
This PR fixes a bug in which there is no note displayed under an individual faculty member in the meeting modal when faculty `notes` is `null.`. This was occurring because the previous code was only checking whether `notes` was equal (`===`) to an empty string and not considering the case when `notes` is equal to `null.` The fix makes it so that if `notes` is equal to `null`, the text 'No Notes' will be displayed under the faculty member.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #331

### DIFF
--- a/src/client/components/pages/Courses/MeetingModal.tsx
+++ b/src/client/components/pages/Courses/MeetingModal.tsx
@@ -157,7 +157,7 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
                     </h4>
                     <p>
                       {
-                        instructor.notes === ''
+                        !instructor.notes || instructor.notes === ''
                           ? <StyledFacultyNote>No Notes</StyledFacultyNote>
                           : instructor.notes
                       }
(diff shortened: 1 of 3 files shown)


src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -157,7 +157,7 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
                     </h4>
                     <p>
                       {
-                        instructor.notes === ''
+                        !instructor.notes || instructor.notes === ''
```
rmainseas, line 160:
```suggestion
                        !instructor.notes
```

Since JavaScript considers `''` and `null` to both be a falsy values, you could probably get away with just the left side of the expression here.
image

rmainseas (changes requested):
Looks great. :+1: 

Just one quick thing.

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Ah my bad! Thanks Robert.

rmainseas (approved):
Looks great. Merge away.

PR #337, opened 2021-05-07
Fix table alias in expression

jonseitz:
This Should fix the issue where new migrations try to re-create the Schedule Entry View.

Previously I had changed the ScheduleEntryView definition so that it would no longer depend on the Room Listing View. In doing so I had introduced a bug by using "c" as the identifier for campus, when that was already being used for "course". While fixing that, I must have made some kind of merge/rebase error that kept "c" in the view definition, but changed it in the migration. So TypeORM keeps seeing that difference and trying to generate a new migration that replaces "c".

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #___


natalynyu (approved):
Looks great! Approved

PR #340, opened 2021-05-26
Restore Parent Id in View

jonseitz:
There were two earlier migrations that both changed the RoomBookingInfoView, and since views are always dropped and recreated only the later migration persisted. This migration restores the changes from the earlier migration on top of the later migration, leaving us with only the combined view definition.

Relevant earlier migrations are:
- 1616098074342-RoomBookingInfoView.ts
- 1617223382839-EditNonClassProperties.ts

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #___


natalynyu (approved):
Approved!

PR #344, opened 2021-06-03
Remove time zone offsets

### LINKED ISSUE #327: Meeting start/end time columns should not use a timezone
The `startTime` and `endTime` columns in our `meeting` table currently use the `time with time zone` type in Postgres. This can lead to potential inconsistencies in our data, as times entered during Daylight Savings Time will have a `-04` offset instead of `-05`. Since there's no date information associated with those times, it's not possible for Postgres to normalize those into a consistent format (In fact the Postgres docs [specifically recommend not using `time with time zone`][pg] for this reason).

Given that, we should change those columns to the `time without time zone` type. This makes more sense for the data anyway, since we're only concerned with the time courses are taking place on campus and we don't need to localize times based on the user's current time zone.

[pg]: https://www.postgresql.org/docs/11/datatype-datetime.html#DATATYPE-TIMEZONES

jonseitz:
This PR changes our database schema to use the plain `TIME` (or `TIME WITHOUT TIME ZONE`) type in Postgres. The use of time zones was leading to some weird behavior when running the code in different environments. Also, the Postgres docs recommend against using `TIME WITH TIME ZONE` because time data without a date is effectively impossible to localize, on account of daylight savings.

To implement that change, I've added a new `PGTime` class in `src/common/util` that helps with formatting and comparison of times. This is a fairly simple implementation that, if helpful in the project, is probably worth expanding to be more robust (i.e., handling custom formats for input/output).

Note that the migration associated with this PR is kind of obnoxious, because all of the views that depend on the startTime/endTime need to be dropped and recreated.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #327


rmainseas (approved):
This looks good. Merge away

PR #349, opened 2021-06-10
Upgrade to Node v14

### LINKED ISSUE #342: Update Node Container to 14
Node 10 is now end of life, so we should move over to Node 14, which is the [recommended LTS version](https://nodejs.org/en/about/releases/).

This should just require a change to the `FROM` lines in each stage of our `Dockerfile`, followed by testing. One feature that I know was added is that [`Array.prototype.flatMap()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) should now be available, so we should no longer need `lodash.flatMap` as a dependency.

For reference: [All the node changelogs](https://github.com/nodejs/node/tree/master/doc/changelogs)

jonseitz:
Node v10 is EOL as of the end of April, so this PR updates our node container and github actions to use Node v14, the new LTS version, and npm v7. This introduced some breaking changes in my local setup that I was able to resolve by also updating the nestjs, typeorm, and pg packages, but that in turn introduced a number of breaking changes in our codebase that needed to be addressed.

I've tried to group similar changes together in the same commits wherever possible, so this PR may be easier to review by commit than by file. For instance, there's a particularly noisy commit (a6c047a) that changes almost all of the swagger decorators.

For local testing, you'll likely need to wipe out and reinstall the node_modules folder. You can also build a local copy of our deployment docker image with `docker build -t local-deploy-course-planner .` to make sure that everything builds correctly.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #342


natalynyu (approved):
Looks great! Approved!

PR #351, opened 2021-06-15
Improve additional parameter handling in customRender

### LINKED ISSUE #257: Refactor customRender function
Our existing [`customRender`](https://github.com/seas-computing/course-planner/blob/develop/src/common/__tests__/utils/customRender.tsx#L21-L32) function expects a messageDispatch function to be passed in as a second argument, which will in turn be passed into our `MessageContext` provider. This design of passing in context values as sequential arguments won't scale well if we want to add additional properties to the custom rendering. After some discussion on slack, the team agreed that we should instead have a `RenderOptions` object as the second argument and pass in our additional properties there. This would:

1. scale better
2. allow us to leverage Typescript hinting
3. let us provide default values to preserve the original `render(<Element />);` syntax if we don't need the additional properties.

- [ ] Create a `RenderOptions` TS interface that enumerates the different values
- [ ] Provide sensible defaults for all values
- [ ] Update existing calls to `render()`
- [ ] Docs

jonseitz:
Just grabbing a tech debt ticket off the backlog. This will improve the way that additional parameters are passed into to our customRender function by accepting an object with name properties as the second argument, instead of a function. It also provides defaults for those parameters, so that it's acceptable to just call `render(<Component />)` without needing to pass an unnecessary stub or empty function.

This required changing a lot of existing calls to `render()`, so I've removed entirely any `dispatchMessage` arguments that weren't being used, and switched to the property value shorthand (i.e. `{ dispatchMessage }` for those that are being used. I've tried to group like changes under the same commits for ease of review.

This is a pretty low priority change, so feel free to take your time on the review.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Low
- [ ] Normal 
- [ ] High 

## Related Issues:

Fixes #257

### DIFF
--- a/src/client/components/general/typography/__tests__/CampusIcon.test.tsx
+++ b/src/client/components/general/typography/__tests__/CampusIcon.test.tsx
@@ -6,16 +6,14 @@ import { CampusIcon } from '..';
 describe('CampusIcon', function () {
   it('renders the first letter on screen', function () {
     const { getByText } = render(
-      <CampusIcon>Allston</CampusIcon>,
-      () => {}
+      <CampusIcon>Allston</CampusIcon>
     );
     const campusStyle = window.getComputedStyle(getByText('A'));
     strictEqual(campusStyle.visibility, 'visible');
   });
   it('Keeps the full text hidden in the DOM content', function () {
     const { container } = render(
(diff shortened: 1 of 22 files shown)


natalynyu (approved):
All looks great! Approved.

PR #364, opened 2021-07-15
Delete Meeting Time - Room Booking Modal (client)

### LINKED ISSUE #318: Room booking modal: Delete meeting time
This ticket provides the functionality to delete a meeting time from the room booking modal and the application state. It does not provide the functionality to alter the DB, this will be done on a different ticket.

* Clicking the 'delete' button will remove the row.
* There is no confirmation pop-up for delete
* Both expanded and closed meeting lines can be deleted.
* Lines can be deleted when another line is expanded.

natalynyu:
This PR adds functionality for the delete buttons for the meeting times in the room booking modal. The `removeMeeting` handler updates the state of all meetings, which lives in the `MeetingModal.` This handler is passed down to the `MeetingTimesList` component.

The tests check that the appropriate meeting is deleted regardless of whether the user is creating or editing meetings and whether the user is deleting the meeting currently being edited, a previously existing meeting, or a newly added meeting. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #318

### DIFF
--- a/src/client/components/pages/Courses/MeetingModal.tsx
+++ b/src/client/components/pages/Courses/MeetingModal.tsx
@@ -314,6 +314,22 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
     }
   };
 
+  /**
+   * A handler to delete a meeting from the current existing meetings of the
+   * course instance. If the deleted meeting was being edited at the time of
+   * deletion, the state of the currently edited meeting is set back to null so
+   * that no meetings are expanded in edit mode.
+   */
+  const removeMeeting = (meeting: CourseInstanceResponseMeeting) => {
+    if (currentEditMeeting !== null) {
+      setCurrentEditMeeting(null);
+    }
+    const updatedMeetings = allMeetings.filter(
+      (currentMeeting) => currentMeeting.id !== meeting.id
+    );
+    setAllMeetings(updatedMeetings);
+  };
+
   return (
     <Modal
       ariaLabelledBy="editMeeting"
@@ -338,6 +354,7 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
                 updateCurrentEditMeeting={updateCurrentEditMeeting}
                 closeCurrentEditMeeting={closeCurrentEditMeeting}
                 showRoomsHandler={searchForRooms}
+                removeMeeting={removeMeeting}
               />
               <h3>
                 Faculty Notes
--- a/src/client/components/pages/Courses/MeetingModal.tsx
+++ b/src/client/components/pages/Courses/MeetingModal.tsx
@@ -340,6 +340,7 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
   const removeMeeting = (meeting: CourseInstanceResponseMeeting) => {
     if (currentEditMeeting && meeting.id === currentEditMeeting.id) {
       setCurrentEditMeeting(null);
+      setShowRoomsData(null);
     }
     const updatedMeetings = allMeetings.filter(
       (currentMeeting) => currentMeeting.id !== meeting.id
--- a/src/client/components/pages/Courses/MeetingModal.tsx
+++ b/src/client/components/pages/Courses/MeetingModal.tsx
@@ -215,6 +215,23 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
   ] = useState('');
 
   /**
+   * Used to create a temporary unique ID for new meetings on the client.
+   * A permanent UUID will be assigned as a result of the server request
+   */
+  const [
+    newMeetingIdNumber,
+    setNewMeetingIdNumber,
+  ] = useState(1);
+
+  /**
+   * Updates the generated meeting id for newly created meetingsby increasing
+   * the current index by 1
+   */
+  const updateNewMeetingIdNumber = (): void => {
+    setNewMeetingIdNumber(newMeetingIdNumber + 1);
+  };
+
+  /**
    * Updates individual fields in the current meeting by merging passed props
    * and values into the object
    */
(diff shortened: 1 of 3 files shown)


src/client/components/pages/Courses/__tests__/MeetingModal.test.tsx
```diff
@@ -933,6 +933,88 @@ describe('Meeting Modal', function () {
...
           });
         });
+        describe('On Delete Behavior', function () {
+          context('when a meeting time is being edited', function () {
+            const cs50FridayMeeting = testCourseInstance[semKey].meetings
+              .filter((meeting) => meeting.day === DAY.FRI)[0];
+            const cs50TuesdayMeeting = testCourseInstance[semKey].meetings
+              .filter((meeting) => meeting.day === DAY.TUE)[0];
+            beforeEach(async function () {
+              const editCS50FridayMeetingButton = await waitForElement(() => document.getElementById('editMeetingButton' + cs50FridayMeeting.id));
```
jonseitz, line 943: Instead of using `document.getElementById`, this should be using something `findByLabelText(/Edit Meeting \d on Friday/)` and querying on the `alt` prop text for the delete button`.

src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -314,6 +314,22 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
...
 
+  /**
+   * A handler to delete a meeting from the current existing meetings of the
+   * course instance. If the deleted meeting was being edited at the time of
+   * deletion, the state of the currently edited meeting is set back to null so
+   * that no meetings are expanded in edit mode.
+   */
+  const removeMeeting = (meeting: CourseInstanceResponseMeeting) => {
+    if (currentEditMeeting !== null) {
+      setCurrentEditMeeting(null);
```
jonseitz, line 325: This also needs to call `setShowRoomsData(null);` to clear the room selection table. Otherwise, if you have a meeting being edit and you've clicked "Show Rooms" to bring up the table before deleting it, the table will persist.

jonseitz (changes requested):
Just two small notes below, and this one that doesn't really fit as an inline comment:

It looks like I introduced a bug by changing the way that the id for the new meeting is generated when you click "Add New Time" -- since the number of meetings will change when one is deleted, we can't rely on `allMeetings.length` to give us a new, unique meeting number. 

As an example, say you have a course instance that has two meetings. If you click "Add New Time", if will create a meeting with id `new-meeting-3` since there will now be three meetings. If you then delete one of the existing meetings -- so there are only two meetings again -- then click "Add New Time Again" if will try to create `new-meeting-3` again and blank out the meeting that was added the first time.

I think the best solution would be to go back to the system you had put in before, of having a `useState` call that creates something like a `nextMeetingIdNumber` and just increment that every time we create a new meeting.

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Oh great point! I did not think of this scenario. I cleared the room selection table in commit 
8a8d427.

src/client/components/pages/Courses/__tests__/MeetingModal.test.tsx
natalynyu: This was fixed in d26877b.

natalynyu:
Thanks for pointing out the issue with the unique IDs of the new meetings - I reverted back to using state, `newMeetingIdNumber`, to make sure that each new meeting ID is unique regardless of which meetings are removed in the process of editing meetings in the modal.

src/client/components/pages/Courses/__tests__/MeetingModal.test.tsx
```diff
@@ -933,6 +933,93 @@ describe('Meeting Modal', function () {
...
+            .filter((meeting) => meeting.day === DAY.TUE)[0];
+          const fridayMeetingIndex = testCourseInstance[semKey].meetings
+            .findIndex(({ id }) => id === cs50FridayMeeting.id);
+          const tuesdayMeetingIndex = testCourseInstance[semKey].meetings
+            .findIndex(({ id }) => id === cs50TuesdayMeeting.id);
+          context('when a meeting time is being edited', function () {
+            beforeEach(async function () {
+              const editCS50FridayMeetingButton = await waitForElement(
+                () => findByLabelText(`Edit Meeting ${fridayMeetingIndex + 1} on ${dayEnumToString(cs50FridayMeeting.day)}, ${cs50FridayMeeting.startTime} to ${cs50FridayMeeting.endTime}${cs50FridayMeeting.room.name === '' ? '' : ` in ${cs50FridayMeeting.room.name}`}`)
+              );
```
jonseitz, line 949: The `waitForElement` call here is redundant, since the `findBy*` queries are asynchronous and can be `await`ed. This could also be an issue down the line as [the `waitForElement` method is set to be deprecated](https://testing-library.com/docs/dom-testing-library/cheatsheet/#async)

And more of a style note, but this text query could be truncated to:

```ts
findByLabelText(`Edit Meeting ${fridayMeetingIndex + 1}`, {exact: false});
```

The additional text is important for adding context to screen readers, but I don't think it's totally necessary for testing purposes. Like I said, that's mostly stylistic, but a shorter query would probably be easier to maintain if we need to update these tests in the future.

jonseitz (changes requested):
Just one more note on some style issues in the tests.

src/client/components/pages/Courses/__tests__/MeetingModal.test.tsx
natalynyu: Ah I see ok thanks for all of this info! I definitely like the shorter query. I updated the tests in 7f6432f.

jonseitz (approved):
Approved!

PR #382, opened 2021-08-03
Meeting Modal: Unsaved Changes Warning

### LINKED ISSUE #371: Room booking modal: State object for unsaved changes
This ticket provides a boolean state object on the room booking modal that is `true` if any changes have been made to any fields in state for the meetings being currently edited, and `false` if no changes have been made.

If any change is made the state will be `true`, even if those changes are later reversed.

natalynyu:
This PR adds the feature of displaying a warning when the user exits the Meeting Modal (either by refreshing the page, clicking outside of the modal, or clicking the cancel button of the modal). When the user makes a change to the meetings state, whether it's by deleting a meeting, adding a new meeting, or editing an existing one, the `isChanged` state is changed from the default `false` to `true.` The value of the `isChanged` state indicates whether a change has been made while in the Meeting Modal. If `isChanged` is true and the user tries exiting the modal, they will receive the unsaved changes warning. By clicking "OK" in the warning, they are agreeing to discard their changes and exit the modal. If they click "Cancel," they can continue making changes in the modal. 

I tried to make the code generic by creating these two functions, `onBeforeUnload` (to handle refreshing) and `onModalClose` (to handle closing the modal). I was thinking of putting these two functions into a `common` directory so that we can use these functions across the app. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #371


src/client/components/pages/Courses/tableFields.tsx
```diff
@@ -143,6 +148,60 @@ export const formatMeetings = (
...
+     * user can continue making changes in the modal.
+     */
+    const onBeforeUnload = (event: Event) => {
+      if (!isChanged) return;
+      event.preventDefault();
+      // It's unclear whether TS will account for browser incompatibility here,
+      // so we use all three methods of setting the confirmation message,
+      // as detailed here:
+      // https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event
+      const legacyEvent = (event as unknown as LegacyEvent);
```
jonseitz, line 176: I'm not totally clear on the need for the `LegacyEvent` -- is that for compatibility with a specific browser?

src/client/components/pages/Courses/tableFields.tsx
```diff
@@ -143,6 +148,60 @@ export const formatMeetings = (
...
+      legacyEvent.returnValue = confirmMessage;
+      return confirmMessage;
+    };
+    window.addEventListener('beforeunload', onBeforeUnload);
+    return () => {
+      window.removeEventListener('beforeunload', onBeforeUnload);
+    };
+  },
```
jonseitz, line 184: Just curious, did you consider using [the `history.block` method from React Router][rrhb]? I haven't explored it too deeply but it does seem to cover some of the edge cases mentioned in that mdn page.

[rrhb]: https://github.com/ReactTraining/history/blob/master/docs/blocking-transitions.md

jonseitz (changes requested):
A couple of notes and questions below for you. Let me know if you want to discuss any of those points further

src/client/components/pages/Courses/tableFields.tsx
natalynyu: Oh yes, so I saw this in the documentation for `beforeunload event`:
image

I was trying to account for the unsaved changes warning showing up in different browsers, so I initially tried to assign a return value property to the `event` by doing...
```
`event.returnValue = confirmMessage;`
```

...however, I got the eslint error: `Assignment to property of function parameter 'event'.` As a result, I created `legacyEvent`:
```
const legacyEvent = event;
legacyEvent.returnValue = confirmMessage;
```
The eslint error I got from the above when trying to assign `confirmMessage` to the `returnValue` property was `Type 'string' is not assignable to type 'boolean',` so this is what prompted me to create `LegacyEvent` as a workaround.

src/client/components/pages/Courses/tableFields.tsx
```diff
@@ -143,6 +148,60 @@ export const formatMeetings = (
...
+      // eslint-disable-next-line no-alert
+      if (window.confirm(confirmMessage)) {
+        setIsChanged(false);
+        setModalVisible(false);
+      }
+    } else {
+      setModalVisible(false);
+    }
+  };
+
```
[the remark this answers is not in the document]
natalynyu: @jonseitz My initial thought was to have this behavior in the `MeetingModal`, but I found that since the `modalVisible` state was in `tableFields` I started putting this logic in the `tableFields` file.

I started trying to move the `onModalClose` and `onBeforeUnload` logic into `MeetingModal` instead - one of the issues I found was that since I was conditionally setting `setModalVisible` to false depending on whether `isChanged` was true and whether the user clicked "OK" on the unsaved changes confirmation, I would need access to the `modalVisible` state within the `MeetingModal.` 

image

It seems like the `modalVisible` state should live in `tableFields,` since clicking an edit button sets the visibility to true and the `isVisible` property checks `modalVisible` to determine whether to show the modal or not. 

I passed down `setModalVisible` to the `Meeting Modal` so that we could control `setModalVisible` in `Meeting Modal`, but then I realized that `onModalClose` would be called `onClose` within `tableFields`:
image

I'm not really sure how to proceed here. Would it be helpful if I pushed up the code I have so far?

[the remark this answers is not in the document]
natalynyu: I found a way to keep all of the behavior in the `MeetingModal` instead of implementing in the parent (5fa8eb3).

src/client/components/pages/Courses/__tests__/MeetingModal.test.tsx
```diff
@@ -1040,17 +1047,137 @@ describe('Meeting Modal', function () {
...
+                  10
+                ),
+                term: meetingTerm,
+              }}
+              onChange={() => { setIsChanged(true); }}
+              onClose={() => {
+                if (!isChanged) {
+                  setOpen(false);
+                }
+              }}
```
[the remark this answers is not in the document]
natalynyu: Good point - These tests have been removed and replaced with end to end tests (6fb3d42).

src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -227,20 +248,66 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
...
       setMeetingModalFocus();
-      setAllMeetings(instanceMeetings);
+      if (!initialized) {
+        setAllMeetings(instanceMeetings);
+        setInitialized(true);
+      }
     } else {
       setCurrentEditMeeting(null);
       setShowRoomsData(null);
+      setInitialized(false);
```
jonseitz, line 258: Rather than create this `initialized` state value, what if we just split the `setAllMeetings(instanceMeetings)` call off into a separate `useEffect` hook that only listens for changes in `isVisible` and `setAllMeetings`? It's not a huge difference, but it might make it a little easier to keep track of what's going to get updated when

src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -227,20 +248,66 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
...
+    const onBeforeUnload = (event: Event) => {
+      if (!isChanged) return;
+      event.preventDefault();
+      // It's unclear whether TS will account for browser incompatibility here,
+      // so we use all three methods of setting the confirmation message,
+      // as detailed here:
+      // https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event
+      const legacyEvent = (event as unknown as LegacyEvent);
+      legacyEvent.returnValue = confirmMessage;
+      return confirmMessage;
```
jonseitz, line 275: I'm still annoyed that Chrome doesn't follow the spec here, but I'll get over it.

One thing that is probably worth changing is that, instead of creating a `LegacyEvent` interface just for this use case, what if we set the `returnValue` to `false` (which is the original type) and return the `confirmMessage` separately. That still seems to work on my end:

```ts
// Need to disable this rule for browser compatibility reasons
// eslint-disable-next-line no-param-reassign
event.returnValue = false;
return confirmMessage;
```

src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -266,6 +333,17 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
...
+  /**
+   * Calls the onChange function to indicate that there has been a change in the
+   * meetings of the current modal. If the user exits the modal without saving,
+   * there will be a warning message to prevent the user from losing changes.
+   */
+  const signalModalChange = () => {
+    if (!isChanged) {
+      setIsChanged(true);
+    }
+  };
```
jonseitz, line 345: Could we just call `setIsChanged(true);` in place of `signalModalChange` on lines 356 and 469 below? I don't think it should trigger any effect hooks to set the state value to the same value....

jonseitz (changes requested):
Just a couple little things to clean up below. They're all pretty minor, but this is already such a complicated component that I think it's worth trying to streamline it as much as possible.

Overall this is looking really great. You've done some nice work here!

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Sure thing! I originally had the `signalModalChange` function to try to have as many reusable parts as possible for the others modals and also to try not to make calls to update the state unless necessary, but this seems simple enough to not be in a function.

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Sure I think that this works - I made the change in commit 274b016.

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Thanks for the suggestion! I updated this in a594352.

src/client/components/pages/Courses/MeetingModal.tsx
```diff
@@ -279,8 +268,19 @@ const MeetingModal: FunctionComponent<MeetingModalProps> = function ({
...
-    initialized,
-    setInitialized,
+  ]);
+
+  /**
+   * Only set the meetings in the modal if the modal is visible and the
+   * meetings haven't already yet been set.
+   */
+  useEffect(() => {
+    setAllMeetings(instanceMeetings);
```
jonseitz, line 278: This still needs to be wrapped in `if (isVisible) { ... }`, otherwise it'll fire off a bunch of state updates on closed modals when the page loads.

jonseitz (changes requested):
Just one small tweak on the last round of changes, but otherwise all looks good.

src/client/components/pages/Courses/MeetingModal.tsx
natalynyu: Oops! Thanks for catching that. This was fixed in 4ad8b38.

jonseitz (approved):
This is all set for liftoff! :rocket:

Issue #358, opened 2021-06-25
Make use of time more consistent


jonseitz:
Based on [a conversation][1] in #356, we need to make sure we're consistently handling times in the same manner. I propose:

- Any _time as data_ should be in 24-hour time, with hours, minutes, and seconds (e.g. `"1:30:00"`)
- Any _time as information_ should be shown in 12-hour AM/PM time, with hours and minutes (e.g. `"1:30 PM"`)
- **For calculations on time, do not use `Date`**, as this will implicitly set a timezone based on the current date (i.e. end up being off by an hour at some point)
- We should have a single class/method for converting back and forth between 12- and 24-hour time.

[1]: https://github.com/seas-computing/course-planner/pull/356#discussion_r658134088

PR #383, opened 2021-08-04
Meeting Modal Save Button

### LINKED ISSUE #367: Room booking modal: Save changes
This ticket provides the functionality to save any changes made using the room booking modal.

On clicking the "Save" button:

* Client side validation will run (already implemented)
* Kicks off PUT request to API endpoint
* Loading spinner is shown whilst request is in process (overlaid over the modal)
* Handle any server side validation errors that come back (server side validation already in place)
* Server side error messages will be displayed, stacked, just above the footer component in the modal. They will utilize the same styles as the other validation error messages in the modal.
  * Error message for room booking conflict will be of the format, "[Room name] is not available at [time and day]"
* On error, user expected to fix then press "save" again.
* On successful save, update list of meetings in main table with returned array of new / amended meetings.
* Close modal

This ticket also implements the PUT request method on the client side.
The request should follow the DTO as defined.

jonseitz:
This PR implements the "Save" functionality on the Meetings Modal for course instances. When the save button is clicked, the modal will run its local validation, then send the data to the server and wait for the response (during which time it will show a spinner and block any other edits). After saving, it will merge the data back into the courses list.

This also sets up end-to-end testing with our existing database mocking by creating an `mockAdapter` that can be loaded into axios and used to run all of our client API queries through supertest's API, allowing us to manipulate the UI with react testing library. This doesn't handle browser testing, since under the hood it's still using jsdom, but it does let us make sure that data is passing back-and-forth between the client and server code successfully.

I did set up a simple version of a modal error to display server errors thrown during the save process, which will need to be replaced when we have a real `ModalError` component to use there instead.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #367


natalynyu (changes requested):
This looks great! It's really cool to see the end to end tests.

One note that I had was when doing manual testing, I found that if I changed the day of the week to one that had conflicts, the row would close, and I would get an error message like so:
image
Maybe it is beyond the scope of our MVP, but I think it would be nice to either print out which room specifically is not available in the error message for clarity in case there are many meetings in the list, or if we could keep the row that caused the error message to remain open. 

Also, I found that when I close the meeting modal with an error message, the error message is still present when I reopen the modal. I think we need to clear out the error message when in the `closeHandler` of the MeetingModal.

natalynyu (commented):
@jonseitz Thanks Jonathan! Sorry this took a while - I fell down a rabbit hole of trying to figure out what was going on with the start and end times.

I found that when `closeCurrentEditMeeting` is called for the first time when an edit meeting button is clicked, the check for whether `validateTimes` initially always passes because at that point, the state values for the start and end time are set to empty strings.

Then in `mergeMeetings`, the `currentEditMeeting` time values are in 12 hour display time.  I bring this up in this comment because I found that if we convert the `currentEditMeeting` times to 24 hour time in `mergeMeetings,` it seems to fix the time validation bug that we saw earlier today (`validateTimes` was comparing 12 hour time strings) :
image

Alternatively, we could call `convert12To24HourTime` inside of the `validateTimes` function. The advantage of this is that whenever `validateTimes` is called, we are definitely comparing 24 hour time strings - perhaps this approach makes mores sense for that reason: 
image

Do you have thoughts on this?

One other little note that I had had to do with the naming of the function `closeCurrentEditMeeting`: I thought that since the primary purpose of the function was to open a new meeting entry that it made more sense to me for it to be called `openCurrentEditMeeting,` especially in the case when the modal initially loads and no particular meeting has yet been selected. However, this is just a preference for my own understandability and not really a big deal.

jonseitz:
@natalynyu Ah, I see. Good sleuthing! I think that ultimately traces back up to #358, where we're going change all of the times in the app to be in 24 hour time (and only convert them for display). Of the two approaches you outline I think adding the conversion to `validateTimes` makes a little more sense. 

That said, my preference would be to just implement the changes necessary for #358 and fix these issues across the board, rather than needing to call `convert12to24HourTime` all over the place.

re: `closeCurrentEditMeeting`, I agree the naming is confusing. My thought process at the time was that `close` made more sense as that's what happened when you called it with no arguments, but in practice I think we are probably using it more to open meetings. What would you think about something like `toggleCurrentEditMeeting`?

natalynyu:
@jonseitz Ok sounds good! Yes I agree - I think that `toggleCurrentEditMeeting` would be a good name to cover all cases.

natalynyu (approved):
LGTM!

Issue #389, opened 2021-09-01
Non-class meetings POST / PUT controller methods


chrxr:
This ticket provides the POST and PUT controller methods for creating and editing non-class meetings.

Creating a new non-class meeting parent also populates blank non-class meeting instances in the same way as if you were creating a new course via the course admin page.

PUT and POST requests will contain the following data:
  * Title (string)
  * Expected size (int)
  * Area (id)
  * Contact name (string)
  * Contact email (email)
  * Contact telephone (tel)
  * Notes (string)
  * ID of non-class parent (for PUT request only)

Only Area and Title are required fields.

PR #390, opened 2021-09-03
Add CreateNonClassParent DTO

rmainseas:
This PR introduces the DTO representing the shape of the data being used to create a non-class parent on the server. Child data such as non-class events are not included in this DTO because those will be generated automatically by the server. Similarly meetings will be added on an as-needed basis by the user using the GUI, so those aren't sent over either.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Delivers the first piece of #389

### DIFF
--- a/src/common/dto/nonClassMeetings/CreateNonClassParent.dto.ts
+++ b/src/common/dto/nonClassMeetings/CreateNonClassParent.dto.ts
@@ -1,0 +1,53 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsUUID } from 'class-validator';
+
+export default abstract class CreateNonClassParentDTO {
+  @ApiProperty({
+    type: 'string',
+    example: 'a49edd11-0f2d-4d8f-9096-a4062955a11a',
+    description: 'Area primary key ID',
+  })
+  @IsUUID()
+  public area: string;
+
+  @ApiProperty({
+    description: 'The title of this non class parent',
+    example: 'Reading group',
+  })
+  public title: string;
+
+  @ApiProperty({
+    description: 'The contact name for this non-class parent',
+    example: 'James Waldo',
+    nullable: true,
+  })
+  public contactName?: string;
+
+  @ApiProperty({
+    description: 'The contact email for this non-class parent',
+    example: 'j.waldo@seas.harvard.edu',
+    nullable: true,
+  })
+  public contactEmail?: string;
+
+  @ApiProperty({
+    description: 'The contact phone number for this non-class parent',
+    example: '(123) 456-7890',
+    nullable: true,
+  })
+  public contactPhone?: string;
+
+  @ApiProperty({
+    description: 'Any misc. notes users wish to record alongside the non class parent',
+    example: 'Only occurs in odd-numbered years',
+    nullable: true,
+  })
+  public notes?: string;
+
+  @ApiProperty({
+    description: 'Non class parent predicted enrollment size',
+    example: 82,
+    nullable: true,
+  })
+  public expectedSize?: number;
+}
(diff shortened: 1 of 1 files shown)


src/common/dto/nonClassMeetings/CreateNonClassParent.dto.ts
```diff
@@ -0,0 +1,53 @@
...
+    nullable: true,
+  })
+  public contactEmail?: string;
+
+  @ApiProperty({
+    description: 'The contact phone number for this non-class parent',
+    example: '(123) 456-7890',
+    nullable: true,
+  })
+  public contactPhone?: string;
```
jonseitz, line 38: As with email above, there is an `@IsPhoneNumber` validator that we can use with `@IsOptional`. It does take a country code as an argument, which we'd want to set to US (though I am _slightly_ concerned that we could have people wanting to use non-US phone numbers here, that seems like an edge case when we're managing physical spaces on a US campus.)

src/common/dto/nonClassMeetings/CreateNonClassParent.dto.ts
```diff
@@ -0,0 +1,53 @@
...
+    nullable: true,
+  })
+  public notes?: string;
+
+  @ApiProperty({
+    description: 'Non class parent predicted enrollment size',
+    example: 82,
+    nullable: true,
+  })
+  public expectedSize?: number;
```
jonseitz, line 52: Should have `@IsInt`and `@IsOptional` validators

jonseitz (changes requested):
Just a couple points below about adding [additional class-validator decorators][decorators] to some of these fields. Even if they're duplicating our type declarations, the validators will perform runtime validation and return richer server errors

[decorators]: https://github.com/typestack/class-validator#validation-decorators

jonseitz (changes requested):
Just a couple points below about adding [additional class-validator decorators][decorators] to some of these fields. Even if they're duplicating our type declarations, the validators will perform runtime validation and return richer server errors

[decorators]: https://github.com/typestack/class-validator#validation-decorators

src/common/dto/nonClassMeetings/CreateNonClassParent.dto.ts
```diff
@@ -0,0 +1,53 @@
...
+    description: 'Area primary key ID',
+  })
+  @IsUUID()
+  public area: string;
+
+  @ApiProperty({
+    description: 'The title of this non class parent',
+    example: 'Reading group',
+  })
+  public title: string;
```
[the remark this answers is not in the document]
rmainseas: You're absolutely right, it should. Fixed in 5ce1671152dbd7199aea9d5a6996bdfe6d161a6e

src/common/dto/nonClassMeetings/CreateNonClassParent.dto.ts
rmainseas: Fixed in 5ce1671152dbd7199aea9d5a6996bdfe6d161a6e

src/common/dto/nonClassMeetings/CreateNonClassParent.dto.ts
rmainseas: Yeah, I think that's somewhat valid but as you mention - it seems extremely unlikely that someone on campus booking a physical space would have a non-US phone number. If this were an app designed for international consumption, perhaps but I agree it seems like a very extreme edge case. If it does come up, we can always loosen the validation to take account of this...

rmainseas: Fixed in Fixed in 5ce1671152dbd7199aea9d5a6996bdfe6d161a6e

jonseitz (approved):
Approved

rmainseas:
Thanks @jonseitz !

PR #394, opened 2021-09-10
Update Font to sans-serif

### LINKED ISSUE #386: Switch to MarkOne sans serif font
Switch all text in the app to use the MarkOne sans serif font. Adjust styling as necessary.

natalynyu:
This PR creates a wrapper span component, `FontWrapper` that updates the font. Each of the table data cells are wrapped so that the font can be updated to sans-serif.
## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #386

### DIFF
--- a/src/client/components/general/typography/FontWrapper.tsx
+++ b/src/client/components/general/typography/FontWrapper.tsx
@@ -1,0 +1,10 @@
+import styled from 'styled-components';
+
+/**
+ * A wrapper span component that converts the font to sans-serif
+ */
+const FontWrapper = styled.span`
+  font-family: "Roboto", sans-serif;
+`;
+
+export default FontWrapper;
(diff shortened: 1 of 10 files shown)


jonseitz (changes requested):
While this is achieving the goal of changing the fontst from monospace to sans-serif, I don't think this is the right approach. Rather then leaving the default table font monospace and overriding it for _every_ cell in the app, I think  we should just change the definition for the `TableCell`/`TableRowHeadingCell` in Mark One to use the `font.body` spec from the mark one theme. If you're OK with taking that approach, I think we could just close this PR and handle this on the mark-one repo instead. What do you think?

natalynyu:
@jonseitz Thanks Jonathan! I opened [PR #115](https://github.com/seas-computing/mark-one/pull/115) in mark-one and assigned you as the reviewer, and I am closing this one.

PR #398, opened 2021-09-13
Feature/387 area ordering

### LINKED ISSUE #387: Ordering of areas in course table
Rows should be ordered in the following way:

1. Alpha by area
2. Alpha by prefix
2. Numeric by number

jonseitz:
This PR updates the API route that populates the Courses tab so that it now returns the list of courses sorted by:

1. area (alphabetically -- e.g. `"AC"` before `"CS"`)
2. prefix (alphabetically -- e.g. `"AC 209A"` before `"CS 50"`)
3. numeric portion of number (numerically -- e.g. `"CS 50"` before `"CS 109A"`)
3. non-numeric portion of number (alphabetically -- e.g. `"CS 109A"` before `"CS 109B"` ) 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #387

### DIFF
--- a/tests/integration/server/courseInstance/courseInstance.service.test.ts
+++ b/tests/integration/server/courseInstance/courseInstance.service.test.ts
@@ -1,4 +1,6 @@
-import { strictEqual, notStrictEqual, deepStrictEqual } from 'assert';
+import {
+  strictEqual, notStrictEqual, deepStrictEqual, ok,
+} from 'assert';
 import { Test, TestingModule } from '@nestjs/testing';
 import { TypeOrmModule, getRepositoryToken, TypeOrmModuleOptions } from '@nestjs/typeorm';
 import { CourseInstanceModule } from 'server/courseInstance/courseInstance.module';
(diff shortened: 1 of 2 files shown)


tests/integration/server/courseInstance/courseInstance.service.test.ts
```diff
@@ -199,6 +201,53 @@ describe('Course Instance Service', function () {
         });
       });
     });
+    describe('Ordering', function () {
+      it('should order by area => prefix => integer => letter', function () {
```
rmainseas, line 205: I would suggest maybe splitting these out into 4 separate tests if possible to make sure that each level of sorting works the way it should on it's own. 

1. Insert some courses into the DB with different areas, make sure the areas are all in the correct order
2. Insert some courses all with the same area, make sure the course prefixes are ordered correctly
3. Insert some courses with the same area and prefix, make sure that the integers are in the correct order
4. Finally, insert some courses with the same area, prefix and integer and make sure the letters are all in in the correct order.

tests/integration/server/courseInstance/courseInstance.service.test.ts
```diff
@@ -199,6 +201,53 @@ describe('Course Instance Service', function () {
...
+        for (let i = 0; i < result.length; i++) {
```
rmainseas, line 206: Would a `forEach` be a better fit here, instead of `for` loop?

rmainseas (changes requested):
Looks great. Just a couple of points regarding the test

tests/integration/server/courseInstance/courseInstance.service.test.ts
jonseitz: I usually prefer to use the `for` loop for cases where we need to compare adjacent values (`result[i]` and `result[i +1]`) since it makes it a little more explicit where each of those values is coming from. I don't think it really makes a difference either way, but did you have a reason in mind for using `forEach`?

tests/integration/server/courseInstance/courseInstance.service.test.ts
rmainseas: Fair enough - that makes sense.

tests/integration/server/courseInstance/courseInstance.service.test.ts
rmainseas: My motivation for suggesting `foreach` was more semantics (i.e: foreach <thing> in <things> rather than indexing `things[i]`). `Array.prototype.foreach` [does provide a second argument to the callback](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#parameters) for the index so you could index adjacent values like that if you wanted to. But I don't feel particularly strongly about it, it was just a passing suggestion.

rmainseas (approved):
Looks good, merge away

PR #400, opened 2021-09-20
Create Non class parent API

rmainseas:
## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Implements part of #389

### DIFF
--- a/src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
+++ b/src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
@@ -1,17 +1,24 @@
 import { ConfigService } from 'server/config/config.service';
 import { TestingModule, Test } from '@nestjs/testing';
 import { stub } from 'sinon';
-import { strictEqual, deepStrictEqual } from 'assert';
-import { year } from 'testData';
+import { strictEqual, deepStrictEqual, fail } from 'assert';
+import { rawAreaList, year, createNonClassParent } from 'testData';
 import {
   computationalModelingofFluidsReadingGroup,
   dataScienceReadingGroup,
 } from 'common/__tests__/data/nonClassEvents';
-import { NonClassEventController } from '../nonClassEvent.controller';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { Area } from 'server/area/area.entity';
 import { NonClassEventService } from '../nonClassEvent.service';
+import { NonClassEventController } from '../nonClassEvent.controller';
 
 const mockNonClassEventService = {
   find: stub(),
+  createWithNonClassEvents: stub(),
+};
+
+const mockAreaRepository = {
+  findOne: stub(),
 };
 
 describe('NonClassEvent controller', function () {
@@ -26,6 +33,10 @@ describe('NonClassEvent controller', function () {
         {
           provide: NonClassEventService,
           useValue: mockNonClassEventService,
+        },
+        {
+          provide: getRepositoryToken(Area),
+          useValue: mockAreaRepository,
         },
       ],
     })
@@ -97,6 +108,33 @@ describe('NonClassEvent controller', function () {
         Object.keys(results),
         [...new Set(mockData.map(({ spring }) => spring.calendarYear))]
       );
+    });
+  });
+  describe('create', function () {
+    it('creates non class parents within an existing area', async function () {
+      const mockArea = rawAreaList[0];
+      mockAreaRepository.findOne.resolves(mockArea);
+
+      await controller.create(createNonClassParent);
+
+      strictEqual(
+        mockNonClassEventService.createWithNonClassEvents.args[0][0].area,
+        mockArea
+      );
+    });
+    it('throws BadRequestException for an invalid area', async function () {
+      mockAreaRepository.findOne.resolves(null);
+
+      try {
+        await controller.create({
+          ...createNonClassParent,
+          area: 'I don\'t exist',
+        });
+        fail('Did not throw');
+      } catch (e) {
+        console.log(e);
+        strictEqual(e.status, 400);
+      }
     });
   });
 });
--- a/src/server/nonClassEvent/nonClassEvent.controller.ts
+++ b/src/server/nonClassEvent/nonClassEvent.controller.ts
@@ -1,11 +1,16 @@
 import {
-  Controller, Get, Inject, Query, UseGuards,
+  BadRequestException,
+  Controller, Get, Inject, Post, Query, UseGuards,
 } from '@nestjs/common';
 import { ConfigService } from 'server/config/config.service';
 import { Authentication } from 'server/auth/authentication.guard';
 import NonClassMeetingResponseDTO from 'common/dto/nonClassMeetings/NonClassMeeting.dto';
-import { NonClassEventService } from './nonClassEvent.service';
+import CreateNonClassParentDTO from 'common/dto/nonClassMeetings/CreateNonClassParent.dto';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Area } from 'server/area/area.entity';
+import { EntityNotFoundError, Repository } from 'typeorm';
 import { NonClassParentView } from './NonClassParentView.entity';
+import { NonClassEventService } from './nonClassEvent.service';
 
 @UseGuards(Authentication)
 @Controller('api/non-class-events')
@@ -15,6 +20,9 @@ export class NonClassEventController {
 
   @Inject(NonClassEventService)
   private service: NonClassEventService;
+
+  @InjectRepository(Area)
+  private areaRepository: Repository<Area>;
 
   /**
    * Retrieves all non class meetings for the specified (or current) academic year.
@@ -48,5 +56,23 @@ export class NonClassEventController {
         }, {}
       );
     return nonClassEvents;
+  }
+
+  @Post('/')
+  public async create(parent: CreateNonClassParentDTO): Promise<any> {
+    const area = await this.areaRepository.findOne({
+      where: {
+        name: parent.area,
+      },
+    });
+
+    if (!area) {
+      throw new BadRequestException(`Cannot create non-class parent for invalid area: ${parent.area}`);
+    }
+
+    await this.service.createWithNonClassEvents({
+      ...parent,
+      area,
+    });
   }
 }
--- a/src/server/nonClassEvent/nonClassEvent.controller.ts
+++ b/src/server/nonClassEvent/nonClassEvent.controller.ts
@@ -8,9 +8,10 @@ import NonClassMeetingResponseDTO from 'common/dto/nonClassMeetings/NonClassMeet
 import CreateNonClassParentDTO from 'common/dto/nonClassMeetings/CreateNonClassParent.dto';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Area } from 'server/area/area.entity';
-import { EntityNotFoundError, Repository } from 'typeorm';
+import { Repository } from 'typeorm';
 import { NonClassParentView } from './NonClassParentView.entity';
 import { NonClassEventService } from './nonClassEvent.service';
+import { NonClassParent } from './nonclassparent.entity';
 
 @UseGuards(Authentication)
 @Controller('api/non-class-events')
@@ -59,7 +60,8 @@ export class NonClassEventController {
   }
 
   @Post('/')
-  public async create(parent: CreateNonClassParentDTO): Promise<any> {
+  public async create(parent: CreateNonClassParentDTO):
+  Promise<NonClassParent> {
     const area = await this.areaRepository.findOne({
       where: {
         name: parent.area,
@@ -70,7 +72,7 @@ export class NonClassEventController {
       throw new BadRequestException(`Cannot create non-class parent for invalid area: ${parent.area}`);
     }
 
-    await this.service.createWithNonClassEvents({
+    return this.service.createWithNonClassEvents({
       ...parent,
       area,
     });
--- a/src/server/nonClassEvent/nonClassEvent.service.ts
+++ b/src/server/nonClassEvent/nonClassEvent.service.ts
@@ -1,18 +1,30 @@
 import { InjectRepository } from '@nestjs/typeorm';
-import { Repository } from 'typeorm';
+import { DeepPartial, Repository } from 'typeorm';
 import { TERM } from 'common/constants';
 import { MeetingListingView } from 'server/meeting/MeetingListingView.entity';
 import { RoomListingView } from 'server/location/RoomListingView.entity';
 import { NonClassParentView } from './NonClassParentView.entity';
 import { NonClassEventView } from './NonClassEvent.view.entity';
+import { Semester } from '../semester/semester.entity';
+import { NonClassParent } from '../nonClassParent/nonclassparent.entity';
+import { NonClassEvent } from './nonclassevent.entity';
 
 export class NonClassEventService {
   @InjectRepository(NonClassParentView)
-  private parentRepository: Repository<NonClassParentView>;
+  private parentViewRepository: Repository<NonClassParentView>;
+
+  @InjectRepository(Semester)
+  private semesterRepository: Repository<Semester>;
+
+  @InjectRepository(NonClassParent)
+  private parentRepository: Repository<NonClassParent>;
+
+  @InjectRepository(NonClassEvent)
+  private eventRepository: Repository<NonClassEvent>;
 
   public async find(calendarYear: number):
   Promise<NonClassParentView[]> {
-    const nonClassEvents = this.parentRepository.createQueryBuilder('p')
+    const nonClassEvents = this.parentViewRepository.createQueryBuilder('p')
       .leftJoinAndMapOne(
         'p.spring',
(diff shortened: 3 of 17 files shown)


src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
```diff
@@ -99,4 +110,31 @@ describe('NonClassEvent controller', function () {
...
+        await controller.create({
+          ...createNonClassParent,
+          area: 'I don\'t exist',
+        });
+        fail('Did not throw');
+      } catch (e) {
+        console.log(e);
+        strictEqual(e.status, 400);
+      }
+    });
```
jonseitz, line 138: You could also do something like:

```ts
assert.rejects(() => controller.create({...}), BadRequestException);
```

src/server/nonClassEvent/nonClassEvent.controller.ts
```diff
@@ -49,4 +60,23 @@ export class NonClassEventController {
       );
     return nonClassEvents;
   }
+
+  @Post('/')
```
jonseitz, line 64: This is missing docs, both in docblock form and in `@Api` tags from swagger describing the body, return values, exceptions, etc.

It should probably have a `RequireGroup` guard restriction, though I don't think we have a group that matches the use case at present -- I _think_ we had talked about letting no-Admin users create their own NonClassEvents that would need to be approved. Maybe we can temporarily add a `GROUP.NON_CLASS` until we have more clarity on that.

src/server/nonClassEvent/nonClassEvent.service.ts
```diff
@@ -56,4 +68,16 @@ export class NonClassEventService {
       .orderBy('p.area', 'ASC');
     return nonClassEvents.getMany();
   }
+
+  public async createWithNonClassEvents(parent: Partial<NonClassParent>):
```
jonseitz, line 72: This is missing docs

jonseitz (changes requested):
Ah, I see what you were saying about the re-exports now. That's really useful, and should definitely save some headaches with other modules. Nicely done.

There is an issue with the actual nonClassEvents that are being created, where they're not being associated with semesters correctly. I've noted that below, along with a few other style cleanup things.

I'm also seeing a bunch of these Nest errors in the test output:

```
In order to use "defaultStrategy", please, ensure to import PassportModule in each place where AuthGuard() is being used. Otherwise, passport won't work correctly.
```

I think the issue might be that the `AuthModule` needs to be registered in the `imports` section of `createTestingModule` in the `NonClassEventService` tests, as it is in the `NonClassEventController` integration tests.

src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
rmainseas: Oh, that's much nicer! Fixed in 8bfffe8e03e002357379e7d3b746558efddd80f4.

I don't think I knew about `assert.rejects`

src/server/nonClassEvent/nonClassEvent.service.ts
```diff
@@ -56,4 +68,16 @@ export class NonClassEventService {
...
+  Promise<NonClassParent> {
+    const semesters = await this.semesterRepository.find({});
+
+    return this.parentRepository.save({
+      ...parent,
+      nonClassEvents: semesters.map((semester) => ({
+        ...this.eventRepository.create(semester),
```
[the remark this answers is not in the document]
rmainseas: Ah yes - well spotted. 

Also, I would have expected that to raise some kind of error from typescript since `.create` expects `DeepPartial<NonClassEvent>` and `semester` is of type `Semester`.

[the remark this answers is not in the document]
rmainseas: Fixed this in 3b726db4bde990c7ccf88d68b4db815f047202a3

src/server/nonClassEvent/nonClassEvent.service.ts
rmainseas: D'oh! So it is. Fixed that in 7b83146d1dd91e2e8df2d0e9065b7b52ebf1bdf7 as well as added some docs for the `find` method above it.

src/server/nonClassEvent/nonClassEvent.controller.ts
rmainseas: > This is missing docs, both in docblock form and in `@Api` tags from swagger describing the body, return values, exceptions, etc.

Well spotted. I've fixed that in ca24bc118e91121ccbdb4f9d13e2ea4bcb427522 as well as updating the `CreateNonClassParentDTO` in ae561bd7523f425e8ac7f3ea6f07bfa8430694e8 to denote the optional properties as such.

> Maybe we can temporarily add a `GROUP.NON_CLASS` until we have more clarity on that.

Added that in 244bcb5559b847d1d3e24f86dcf1cc4b73331221 (despite the commit message being a little off - I originally set it to `GROUP.ADMIN` and then amended it before pushing)

src/server/nonClassEvent/nonClassEvent.controller.ts
```diff
@@ -49,4 +60,23 @@ export class NonClassEventController {
...
+  Promise<NonClassParent> {
+    const area = await this.areaRepository.findOne({
+      where: {
+        name: parent.area,
+      },
+    });
+
+    if (!area) {
+      throw new BadRequestException(`Cannot create non-class parent for invalid area: ${parent.area}`);
+    }
```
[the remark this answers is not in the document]
rmainseas: This was the approach I originally took, but doing this requires most of the method body to be wrapped in the `try{}`....

src/server/nonClassEvent/nonClassEvent.controller.ts
rmainseas: Also updated my tests in 9e3a0707f8a02ca5a5ad3f98a79f186bf7aec130 to force `authStub` to return a user with the correct group membership to edit non-class events(otherwise, the tests fail).

This also involved the creation of said user in our test data, so we have that going forward.

tests/integration/server/nonclassevents/nonclassevent.service.test.ts
```diff
@@ -137,4 +146,46 @@ describe('NonClassEvent Service', function () {
...
+      const area = await areaRepository.findOne();
+      const semesterCount = await semesterRepository.count();
+
+      const parent = await service.createWithNonClassEvents({
+        area,
+        title: appliedMathematicsReadingGroup.title,
+      });
+
+      strictEqual(parent.nonClassEvents.length, semesterCount);
+    });
```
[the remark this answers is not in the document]
rmainseas: Agreed. Fixed in cf27e152d89abe1c8948ea8ecfd58caf6cb62ea4

src/server/nonClassEvent/nonClassEvent.controller.ts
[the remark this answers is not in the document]
rmainseas: So, I was originally going to push back on this out of concern that wrapping this in a try/catch might inadvertently swallow exceptions from elsewhere.

However, it doesn't look like that's the case. But I've added a test case for that to be sure. 

Updated all this in 7d529164619b11e64356d6bac4421070365c22e7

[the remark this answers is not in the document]
rmainseas: So, I was originally going to push back on this out of concern that wrapping this in a try/catch might inadvertently swallow exceptions from elsewhere.

However, it doesn't look like that's the case. But I've added a test case for that to be sure. 

Updated all this in 7d529164619b11e64356d6bac4421070365c22e7

src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
```diff
@@ -99,4 +131,44 @@ describe('NonClassEvent controller', function () {
       );
     });
   });
+  describe('create', function () {
+    it('creates non class parents within an existing area', async function () {
+      const mockArea = rawAreaList[0];
+      mockAreaRepository.findOneOrFail.resolves(mockArea);
+
+      await controller.create(createNonClassParent);
```
jonseitz, line 139: I was testing this locally with `throw e` added to the catch block of `createWithNonClassEvents` and this test started failing with an error about not being able to destructure `id` -- I assume that's happening because `mockNonClassEventService.createWithNonClassEvents` is just a `stub()`, so it's not actually returning any value to destructure.

src/server/nonClassEvent/nonClassEvent.controller.ts
```diff
@@ -49,4 +79,37 @@ export class NonClassEventController {
...
+    required: true,
+  })
+  public async create(@Body() { area, ...parent }: CreateNonClassParentDTO):
+  Promise<NonClassParentResponse> {
+    try {
+      const dbArea = await this.areaRepository.findOneOrFail({
+        where: {
+          name: area,
+        },
+      });
```
jonseitz, line 103: The `CreateNonClassParentDTO` says that the `area` field must be a `UUID`, so I expect that this is always going to fail.

tests/integration/server/nonclassevents/nonclassevent.controller.test.ts
```diff
@@ -107,4 +127,60 @@ describe('Non Class Event API', function () {
...
+            false
+          );
+        });
+      });
+      describe('User is authorized', function () {
+        beforeEach(function () {
+          authStub.resolves(nonClassEventManager);
+        });
+        it('creates non-class parent data', async function () {
+          mockAreaRepository.findOneOrFail.resolves(rawAreaList[0]);
```
jonseitz, line 170: I don't think `rawAreaList` is the right data to use here, as it only returns a `name` field instead of a full `Area` entity.

Additionally, it seems that none of these tests are actually testing that `areaRepository.findOneOrFail` is actually searching correctly with the data in the request -- Anywhere the controller route is invoked the method is just being stubbed out to return data. I think that's created a blind spot in our tests, as it's not picking up that the controller code is searching by name, while the DTO is enforcing that `area` should be the id of an area. One solution would be to combine these `service` and `controller` integration tests, and test everything against the database using only the API method. (That's what I've taken to doing, e.g. the [Instructor PUT method][1])

[1]: https://github.com/seas-computing/course-planner/blob/develop/tests/integration/server/courseInstance/courseInstance.controller.test.ts#L584

jonseitz (changes requested):
A few more fixes noted below, but overall this is shaping up nicely

tests/integration/server/nonclassevents/nonclassevent.controller.test.ts
jonseitz: Just wanted to clarify this comment from Friday: I'm not recommending rewriting all the tests to run everything through the API, just highlighting it as a strategy for future tests.

In this case I think we just need to:

- [ ] Return a full `Area` entity
- [ ] Correct the shape of the data to match the DTO (or vice versa)

src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
```diff
@@ -99,4 +131,44 @@ describe('NonClassEvent controller', function () {
...
+      void rejects(() => controller.create({
+        ...createNonClassParent,
+        area: 'I don\'t exist',
+      }), BadRequestException);
+    });
+    it('allows other errors to bubble', function () {
+      mockAreaRepository.findOneOrFail.rejects(Error);
+
+      void rejects(() => controller.create(createNonClassParent), Error);
+    });
```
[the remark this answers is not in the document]
rmainseas: Ah, that's what I was after, thanks! I was getting typescript warnings about the rejection. 

Fixed in c5ac417a7c3d4a8d6df9561b6dbc508c8184d3aa

src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
```diff
@@ -99,4 +131,44 @@ describe('NonClassEvent controller', function () {
...
+    it('throws BadRequestException for an invalid area', function () {
+      mockAreaRepository.findOneOrFail.rejects(EntityNotFoundError);
+
+      void rejects(() => controller.create({
+        ...createNonClassParent,
+        area: 'I don\'t exist',
+      }), BadRequestException);
+    });
+    it('allows other errors to bubble', function () {
+      mockAreaRepository.findOneOrFail.rejects(Error);
```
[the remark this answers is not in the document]
rmainseas: Fixed in 15c74630d1b81f660c3a0b5c8ed8726c7b5ac5f1

src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
```diff
@@ -99,4 +131,44 @@ describe('NonClassEvent controller', function () {
...
+      mockParentRepository.findOne.resolves(nonClassParent);
+      const parent = await controller.create(createNonClassParent);
+
+      strictEqual(
+        parent,
+        nonClassParent
+      );
+    });
+    it('throws BadRequestException for an invalid area', function () {
+      mockAreaRepository.findOneOrFail.rejects(EntityNotFoundError);
```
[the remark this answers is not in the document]
rmainseas: Ah, right. Yes - well spotted!

Fixed that in 15c74630d1b81f660c3a0b5c8ed8726c7b5ac5f1

src/server/nonClassEvent/nonClassEvent.controller.ts
```diff
@@ -62,21 +81,35 @@ export class NonClassEventController {
...
+      const { id } = await this.service.createWithNonClassEvents({
+        ...parent,
+        area: dbArea,
+      });
+      return this.parentRepository.findOne(id);
+    } catch (e) {
+      if (e instanceof EntityNotFoundError) {
```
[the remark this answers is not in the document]
rmainseas: Fixed in a570cff4e491efa05c09bc558ea6bbdc00ac813f.

Also added a `jsError` to our test data for usage elsewhere. Thought that might be useful.

src/common/dto/nonClassMeetings/NonClassParentResponse.dto.ts
```diff
@@ -0,0 +1,59 @@
...
+  public contactPhone?: string;
+
+  @ApiProperty({
+    example: 'Some notes',
+    nullable: true,
+  })
+  public notes?: string;
+
+  @ApiProperty({
+    example: 'e77babce-1a85-4d5f-9d24-1c581b6bf6bb',
```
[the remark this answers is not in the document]
rmainseas: D'oh! Yes, it should. That's what I get for copy/pasting.

Fixed in 0a7f384b67df4066801245b24c8533b928aad21b

src/server/nonClassEvent/__tests__/nonClassEvent.controller.test.ts
rmainseas: Ah, yes - that probably also comes back to what you said about the catch block swallowing other errors.

Fixed in 98aec625686fb3acfb30afb7cd073ada0590dbcc

src/server/nonClassEvent/nonClassEvent.controller.ts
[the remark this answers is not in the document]
jonseitz: FYI we already have an `error` value in our test data: https://github.com/seas-computing/course-planner/blob/main/src/common/__tests__/data/generics.ts#L52

src/server/nonClassEvent/nonClassEvent.controller.ts
rmainseas: Ah, yes - good catch, I've updated this to search using `area` as the PK id(which is what it was supposed to be in the first place)

tests/integration/server/nonclassevents/nonclassevent.controller.test.ts
rmainseas: I've updated the repository to now resolve to an instance of `Area` and corrected the way the controller behaves to treat the incoming `area` value as a UUID instead of the area name. 

Both of these were addressed in 86ca9acddef9a66a15a0f8b33583a68af058f18d

src/common/__tests__/data/area.ts
```diff
@@ -16,3 +18,11 @@ export const rawAreaList = [
...
   { name: 'SEM' },
 ];
+
+/**
+ * An example of an [[Area]] entity
+ */
+export const appliedMath = Object.assign(new Area(), {
+  id: '4d4e622a-a843-43e1-8f53-c99532583178',
+  name: 'AM',
+} as Partial<Area>);
```
jonseitz, line 28: Did a quick test on my end and it doesn't seem like this needs to be converted to `Partial<Area>`, I'd rather avoid unnecessary type conversions

src/common/__tests__/data/errors.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * An instance of the `Error` class built in to JavasScript
+ */
+export const jsError = new Error('Plain JavaScript Error');
```
jonseitz, line 4: I mentioned this is a reply to another comment,  but we do have an `error` already defined in our test data in `generics.ts`

PR #402, opened 2021-09-23
Attempt to stabilize tests

jonseitz:
After running up against yet another string of phantom test failures I decided it was worth trying to make some changes to the test setup. This PR rolls up a couple of small things that, hopefully, should resolve the deadlock issues we've occasionally seen:

1. Instead of running a fresh container for each test suite, we're just going to start a single container, which we can dump and repopulate before each test.
2. The container will not run if passing `--client` or `--server`, i.e., `npm run test -- --client`
3. The container can be skipped by changing the value of `MOCHA_NEEDS_DATABASE` when running the tests
4. The container now uses a random port in the 3xxxx range to avoid port conflicts
5. The container uses a unique name to avoid naming conflicts
6. The container only listens on `127.0.0.1` instead of `0.0.0.0`

(There are more details on each of those changes in the commit messages)

Also, it's not actually connected but I fixed the `SIGINT` behavior so it now appropriately stops the container and exits the test run when you press `CTRL-C`

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #___

### DIFF
--- a/.mochainit.ts
+++ b/.mochainit.ts
@@ -1,14 +1,50 @@
 import { restore } from 'sinon';
+import MockDB from './tests/mocks/database/MockDB';
 
-before(function () {
+before(function (done) {
   process.env.NODE_ENV = 'testing';
   process.env.SERVER_URL = '';
+  if (process.env.MOCHA_NEEDS_DATABASE === '1') {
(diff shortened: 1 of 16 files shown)


natalynyu (approved):
Everything looks great! Approved!

Issue #370, opened 2021-07-19
Generic warning function when closing a modal with unsaved changes


chrxr:
This ticket implements the warning when a user tries to exit a modal with unsaved changes. It should also trigger if the user presses the browser back button or otherwise try to navigate away from the page (https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload).

Warning does not appear if no details within the modal have been changed, i.e. the state for that particular meeting is unchanged.

This function will utilize window.confirm() to display a box with an 'Ok' or 'Cancel' button, and the following text:

"You have unsaved changes. Click 'OK' to discard changes, or 'Cancel' to continue editing."

Function will need to check a state object that confirms whether something has changed or not.

Clicking 'OK' closes the modal, discarding any changes.

Clicking 'Cancel' dismisses the message box and returns the user to the modal.

Whilst the message box is present, the modal is inactive.

PR #410, opened 2021-10-01
Initial implementation of Instructor Modal

### LINKED ISSUE #376: Instructor Modal: Initial view of faculty in modal
* Clicking the edit button in the Intructor column opens an instance of the Mark One modal.
* A state object is created in that modal with a list of the instructors.
* This list is passed down to a table with two columns, which displays the instructor names in the left column, in the order that they are displayed in the courses table.
* The modal header component will have the text: "Edit [course]", e.g. "Edit AC 297r". 
* Save and Cancel buttons will be present in the modal footer component.

jonseitz:
This creates the modal for editing the instructor list assigned to a course instance. At present, it just renders the list, in order, with nonfunctional delete and up/down buttons that will be wired up in future PRs.

Note that I haven't implemented anything related to the unsaved changes  yet, as there's no editing functionality with which to test it.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #376

### DIFF
--- a/src/client/components/pages/Courses/InstructorModal.tsx
+++ b/src/client/components/pages/Courses/InstructorModal.tsx
@@ -1,5 +1,9 @@
 import React, {
-  FunctionComponent, ReactElement, useState, useEffect,
+  FunctionComponent,
+  ReactElement,
+  useState,
+  useEffect,
+  useRef,
 } from 'react';
 import {
   Modal,
@@ -81,6 +85,17 @@ const InstructorModal: FunctionComponent<InstructorModalProps> = ({
   } = currentCourse;
 
   /**
+   * Ref to attach to the internal modal header
+   */
+  const modalHeaderRef = useRef<HTMLHeadingElement>(null);
+
+  /**
+   * Shift the focus to the modal header when it appears on the page
+   */
+  const setMeetingModalFocus = (): void => {
+    setTimeout(() => modalHeaderRef.current?.focus());
+  };
+  /**
    * Keep a local copy of the instructors that we can modify before committing
    * to the server
    */
@@ -96,6 +111,7 @@ const InstructorModal: FunctionComponent<InstructorModalProps> = ({
   useEffect(() => {
     if (isVisible) {
       setAllInstructors(instanceInstructors);
+      setMeetingModalFocus();
     }
   }, [
     isVisible,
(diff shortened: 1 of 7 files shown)


src/client/components/pages/Courses/__tests__/InstructorModal.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+        deepStrictEqual(renderedNames, instructorNames);
+      });
+      it('should render a delete button', function () {
+        const deleteButton = renderResult.queryByLabelText(
+          `Remove ${instructorNames[0]} from ${testCourse.catalogNumber}`,
+          { exact: false }
+        );
+        notStrictEqual(deleteButton, null);
+      });
+      it('Should not render any arrow buttons', function () {
```
natalynyu, line 81: It looks like it does render arrow buttons that are just have no click action/`VARIANT.DEFAULT` colored.

natalynyu (changes requested):
This looks great! I really liked the way you handle the logic of rendering the arrows based off of the indexes and the comprehensiveness of the tests. I just had a few small comments on accessibility and a test description.

I also noticed that the mockup doesn't show the `VARIANT.DEFAULT` buttons. I'm thinking that your approach may be better, since the alt text of the `VARIANT.DEFAULT` button describes that the instructor can't be moved, whereas the mockup doesn't have the arrow at all if the instructor can't be moved in that direction. Just wanted to bring this up as a point in case you have further thoughts on that.

Also I noticed that the ticket #376 says that this will include a state to track whether changes have been made, but maybe this makes more sense to be done later after #370 is merged.

src/client/components/pages/Courses/InstructorModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+    <Modal
+      ariaLabelledBy="edit-instructors-header"
+      closeHandler={closeModal}
+      isVisible={isVisible}
+    >
+      <ModalHeader>
+        <span id="edit-instructors-header">
+          {`Edit Instructors for ${instanceIdentifier}`}
+        </span>
+      </ModalHeader>
```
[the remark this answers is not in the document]
jonseitz: Excellent point, those are missing. Added them in 872f695

src/client/components/pages/Courses/__tests__/InstructorModal.test.tsx
jonseitz: yeah, those should be cleared. Changed the test descriptions in

natalynyu (approved):
Approved!

PR #442, opened 2021-12-02
Course Instance Table Dropdown Filters

### LINKED ISSUE #425: Implement drop-down filtering table headers on Course Instance table
Implement drop-down filters on the following columns in the Course Instance Table:

* Area
* Is SEAS
* Offered

Filtering on offered column should use AND filtering between the two semesters.

natalynyu:
This adds the `area,` `isSEAS,` and `offered` dropdown filters for the course instance table. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #425


src/client/components/pages/Courses/CourseInstanceTable.tsx
```diff
@@ -67,6 +80,62 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
...
+    }
+    if (springOfferedValue !== 'All') {
+      courses = listFilter(
+        courses,
+        { field: 'course.spring.offered', value: springOfferedValue, exact: true }
+      );
+    }
+    return courses;
+  };
+
```
jonseitz, line 138: I think all of this filtering logic needs to be moved up a level, into the `CoursesPage` component. The idea is that we want the `CourseInstanceTable` to just be concerned with how the table itself gets rendered, while the parent component can handle the fetching and manipulation of the data.

Doing that would mean that we'd need to pass down handlers for filtering the different fields. So I think we should also simplify the structure of our state by consolidating all the filter fields into a `filters` state, e.g.

```ts

interface SemesterFilterState {
  offered: OFFERED | 'all';
  // and so on
}

interface FilterState {
  area: string;
  isSEAS: IS_SEAS | 'all';
  spring: SemesterFilterState;
  fall: SemesterFilterState;
  // and so on
}

const [filters, setFilters] = useState<FilterState>({
  area: 'all',
  isSEAS: 'all',
  spring: {
    offered: 'all',
  },
  fall: {
    offered: 'all'
  },
});
```

Then we can use the [functional update][1] argument to `setFilters` to create a generic handler that can be passed down. The basic look of this would be something like this, though you'd need to account for the nested semester field: 

```jsx
const genericFilterUpdate = (field, value) => {
  setFilters((filters) => ({
    ...filters,
    [field]: value,
  }));
};

/* In our Dropdown component */
<Dropdown
  id="area"
  onChange={(evt) => { genericFilterUpdate('area', evt.target.value); }}
/>
<Dropdown
  id="isSEAS"
  onChange={(evt) => { genericFilterUpdate('isSEAS', evt.target.value); }}
/>
```

(That might be tricky to type out the function signature in Typescript, you'd probably need to use some combination of [generics][2], [overloads][3], and [`keyof` types][4])

[1]: https://reactjs.org/docs/hooks-reference.html#functional-updates
[2]: https://www.typescriptlang.org/docs/handbook/2/functions.html#generic-functions
[3]: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads
[4]: https://www.typescriptlang.org/docs/handbook/2/keyof-types.html

src/client/components/pages/Courses/CourseInstanceTable.tsx
```diff
@@ -182,6 +256,117 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
           <TableRow>
```
jonseitz, line 256: As submitted, this is broken because this `<TableRow>` is only rendered when the enrollment fields are included, which is not true in the default view. Since the filters are always going to appear in the same row as the enrollment fields and we'll always have `area` and `course` columns with filters, we can probably remove all checks for `firstEnrollmentField > -1`

src/client/components/pages/Courses/CourseInstanceTable.tsx
```diff
@@ -182,6 +256,117 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
...
+                              setFallOfferedValue(event.currentTarget.value);
+                            } else {
+                              setSpringOfferedValue(event.currentTarget.value);
+                            }
+                          }
+                        }
+                      />
+                    </TableHeadingCell>
+                  );
+                }
```
jonseitz, line 369: This is sort of the inverse of the note above about pulling the filter state up into the `CoursesPage` -- Here I think we want to psuh these individual filter details down to the `tableFields` file, similar to how in the table body we've pushed the rendering of the different cell contents behind the `getValue` function. 

I think what we'd need to do is go through the `tableFields` array and add something like a `getFilter` function to each entry, where you'd basically pass in the `filters` object and the `genericFilterUpdate` method described above. Then this whole block could be replaced by adding:

```tsx
  return field.getFilter(filters, genericFilterUpdate);
```

in place of the `return null` on line 380, after the `if (field.viewColumn === COURSE_TABLE_COLUMN.ENROLLMENT)` block

The main benefit would be to reduce the complexity of the layout by removing the need for all of these `if` blocks. Almost all of our fields will have filters, and for those that don't we can always just have `getFilter` return `null`. 

The other benefit is that we can potentially break things down further an reuse more of the logic, i.e. having `getFilter` call different functions to format dropdowns vs. format text fields.

src/client/components/pages/Courses/CourseInstanceTable.tsx
```diff
@@ -200,51 +385,57 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
         )}
       </TableHead>
       <TableBody>
-        {courseList.map(
+        {filteredCourses(courseList.map(
```
jonseitz, line 388: As noted above, I think this filtering should be done in the `CoursesPage` component, instead of inside the `CourseInstanceTable`. That might bring back the error you were seeing about hooks being called a different number of times, so we may need to look into that a bit more...

jonseitz (changes requested):
A couple of fairly broad comments below, but to summarize the key points:

1. As submitted, this implementation is broken because it only shows the filters when the enrollment columns are present, which is not the case in the default view. That said, when I did add the enrollments the filters themselves are working correctly!
2. We should try to keep the `CourseInstanceTable` as "dumb" as possible, by pulling the filtering code up into the `CoursesPage` and pushing the logic for rendering individual filters into the `tableFields` file.

Overall though, this is a solid filtering implementation and you've got the core functionality working as it should. The suggestions I have for no. 2 above are really aimed at reducing complexity and improving maintainability. I'm happy to hop on a zoom call, or even meet in person (:mask:), to walk through some of these in more detail, because I do think there are going to be some tricky Typescript issues that will come up from this...

src/client/components/pages/Courses/CourseInstanceTable.tsx
natalynyu: Thanks for your help with this! The filtering logic has been moved to the `CoursesPage` level now so that we can pass down the `filteredCourses` to the `CourseInstanceTable.`

src/client/components/pages/Courses/CourseInstanceTable.tsx
natalynyu: This is fixed now so that the filters show whether or not the enrollment columns are showing.

src/client/components/pages/Courses/CourseInstanceTable.tsx
natalynyu: Thanks for your help with this! There is now a `getFilter` function in `tableFields` that generates a dropdown filter for the corresponding column.

jonseitz (approved):
This looks great! Sorry that it turned into a much bigger piece of work than we were initially expecting, but I do think the changes have put us a good place for getting the rest of the filtering functionality squared away, and it's helped resolve some outstanding technical debt (ie. removing the multiple modals). Nice job!

PR #446, opened 2021-12-09
Feature/443 performance

### LINKED ISSUE #443: Performance analysis of models in course instance table.

jonseitz:
The primary change here is adding a few calls to `useMemo` through the `tableFields.tsx` file, which prevent a number of the cells from re-rendering unless their props change. The penalty of re-rendering seems to be most pronounced with the Instructor and Meeting cells, but I've added `useMemo` around all of the cells as a bit of future-proofing. 

Doing some quick testing with React Devtools add-on in Firefox:

- On `develop`, opening or closing a modal typically took around 1500ms of rendering time (+/- 100ms)
- With these changes, opening or closing takes about 450ms (+/- 50ms)

That's enough of an improvement to make the system feel less "broken", but it's still enough to feel laggy. It also only improves the render time after the initial render -- the initial page load still takes the full amount of rendering time. 

To gain further improvements, we'd need to reduce the number of rows rendered in the table. Even though blocking unchanged cells from re-rendering, it seems there is still something of a performance penalty for even checking the memoization. The tests above were done with a table containing 477 rows, so as an experiment I re-ran the profiling with these changes while also arbitrarily cutting down the number of rows:

- With 100 rows, opening or closing a modal take about 100ms (+/- 20ms)
- With 50 rows, opening or closing a modal takes about 70ms (+/- 20ms)
- With 10 rows, opening or closing a modal takes about 30ms (+/- 10ms)

From a user standpoint, I think 50 is right about the sweet spot. But as we've discussed, we'll hold off on doing any further optimization until post-MVP.

(It's also worth noting that filtering the table will also reduce the number of rows that need to be rendered/evaluated/re-rendered, so if a user is only looking at CS courses, the table will probably be a bit smoother too).

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #443

### DIFF
--- a/src/client/components/pages/Courses/__tests__/tableFields.test.tsx
+++ b/src/client/components/pages/Courses/__tests__/tableFields.test.tsx
@@ -1,7 +1,9 @@
 import {
   waitForElement,
 } from '@testing-library/react';
-import React, { FunctionComponent, ReactElement } from 'react';
+import React, {
+  FunctionComponent, ReactElement,
+} from 'react';
 import { spy, SinonSpy } from 'sinon';
 import {
   cs50CourseInstance, ac209aCourseInstance, ac209aCourseInstanceWithoutRooms,
@@ -19,8 +21,10 @@ import {
   formatInstructors,
   formatMeetings,
   formatFacultyNotes,
+  CourseInstanceListColumn,
 } from '../tableFields';
 import { PGTime } from '../../../../../common/utils/PGTime';
+import CourseInstanceResponseDTO from '../../../../../common/dto/courses/CourseInstanceResponse';
 
 describe('tableFields', function () {
   let updateSpy: SinonSpy;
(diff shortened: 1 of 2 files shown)


src/client/components/pages/Courses/__tests__/tableFields.test.tsx
```diff
@@ -29,73 +33,108 @@ describe('tableFields', function () {
...
+          {getValue(course)}
+        </p>
+      );
       it('should return a function to get course-level fields', function () {
-        const getValue = retrieveValue('catalogNumber');
-        strictEqual(
-          getValue(cs50CourseInstance),
+        const { getByText } = render(
+          <TestComponent
+            getValue={retrieveValue('catalogNumber')}
```
rmainseas, line 51: Can we use `COURSE_TABLE_COLUMN.CATALOG_NUMBER` for this (and all the others)?

rmainseas (changes requested):
This looks great to me. Just a quick query regarding the constant/string value.

jonseitz:
Closing this in favor of #454

PR #449, opened 2022-01-03
Allow user to edit default view

### LINKED ISSUE #439: Course instance table: Allow user to create a new view

rmainseas:
## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #439


natalynyu (changes requested):
This is looking good! One thing that I noticed that I didn't include in the code comments is that when you add and remove the instructor columns, the "inconsistent number of hooks" error appears: 
image

This issue should be fixed by the [PR](https://github.com/seas-computing/course-planner/pull/442) that was just merged into `develop` today. That PR should fix this inconsistent number of hooks error by removing the React hooks from the `formatInstructors` function. (This function was being called on every row, and when rows were removed via filtering, the function wasn't being called the same number of times. The workaround was to remove the hooks from the function and to pass down what was needed to the function instead through props). We should check to make sure this error isn't still appearing after `develop` is merged into this branch. Everything else seems to be working great!

src/client/components/pages/Courses/ViewModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+   * Handler to be invoked when the modal closes
+   */
+  onClose: () => void;
+}
+
+/**
+* Displays a list of the instructors associated with a course instance and
+* provides way to add, remove and rearrange them
+*/
+const ViewModal: FunctionComponent<ViewModalProps> = ({
```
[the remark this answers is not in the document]
rmainseas: Yes it was. I copy/pasted but missed the comment. Fixed in f2693bc904f9ebb426711998859ee621cadaa2e7

src/client/components/pages/Courses/SemesterTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * Handler to be invoked when a checkbox is checked or unchecked
+  */
+  onChange: (cols: COURSE_TABLE_COLUMN) => void;
+}
+
+/**
+ * Component representing the list of CourseInstances in a given Academic year
+ */
```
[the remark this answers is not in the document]
rmainseas: Fixed in f2693bc904f9ebb426711998859ee621cadaa2e7

src/client/components/pages/Courses/SemesterTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+  TableBody,
+  Form,
+} from 'mark-one';
+import { COURSE_TABLE_COLUMN, COURSE_TABLE_COLUMN_GROUP, MANDATORY_COLUMNS } from 'common/constants';
+import { ModalFieldsColumn } from './modalFields';
+
+interface SemesterTableProps {
+  columns: ModalFieldsColumn[],
+
+  checked: COURSE_TABLE_COLUMN[],
```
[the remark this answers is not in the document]
rmainseas: Fixed in 4a696e7f4fc0d864056e02da140069b152657808

src/client/components/pages/Courses/ViewModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Ref to attach to the internal modal header
+   */
+  const modalHeaderRef = useRef<HTMLHeadingElement>(null);
+
+  return (
+    <Modal
+      ariaLabelledBy="view-modal-header"
+      isVisible={isVisible}
+    >
```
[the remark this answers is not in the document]
rmainseas: Ah, yes it is - fixed in 4a79d56b11d062fb1a2814e9ccc7493a8b0bcc29.

src/client/components/pages/Courses/CoursesPage.tsx
```diff
@@ -138,20 +145,30 @@ const CoursesPage: FunctionComponent = (): ReactElement => {
...
-              <Dropdown
-                id="select-view-dropdown"
-                name="select-view-dropdown"
-                onChange={(event: React.ChangeEvent<HTMLSelectElement>) => {
-                  setCurrentViewId(event.currentTarget.value);
+              <ViewModal
+                isVisible={viewModalVisible}
+                onClose={() => {
+                  setViewModalVisible(false);
+                }}
```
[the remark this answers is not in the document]
rmainseas: Fixed in ee8a81b0480b59d31767f84b4e639775accc3cbd

src/client/components/pages/Courses/ViewModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }, [isVisible]);
+
+  return (
+    <Modal
+      ariaLabelledBy="view-modal-header"
+      isVisible={isVisible}
+      closeHandler={() => {
+        onClose();
+      }}
```
natalynyu, line 54: @rmainseas Just to confirm, this would allow the selections to be saved without clicking the "Done" button. Is this the intended behavior? I thought that the "done" button was acting as a "save" button. However, it seems like we intend the columns selected to change in the background as you select them in the modal, so it seems like this current implementation is ok (just wanted to confirm with you though).

src/client/components/pages/Courses/SemesterTable.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Handler to be invoked when a checkbox is checked or unchecked
+  */
+  onChange: (cols: COURSE_TABLE_COLUMN) => void;
+}
+
+/**
+ * Component to show/hide columns in the CourseInstance cable
```
natalynyu, line 39: I think that you meant to say "table" instead of "cable" here.

rmainseas:
@natalynyu yeah - I believe that was the decision we came to since having the changes only commit when you press "Done" would require maintaining two pieces of state (and thus additional complexity).

natalynyu (commented):
Looks good! Just wanted to confirm the onClose behavior when the user clicks outside the modal.

src/client/components/pages/Courses/ViewModal.tsx
rmainseas: Yep. That's intentional - I believe that was the decision we came to since having the changes only commit when you press "Done" would require maintaining two pieces of separate state (and thus additional complexity). If someone really gets themselves into a mess and somehow hides everything or whatever, the changes aren't persisted outside of react state - so they can just reload the page and "un-stick" themselves (so-to-speak).

src/client/components/pages/Courses/SemesterTable.tsx
rmainseas: Yes....yes I did. :laughing: 

Fixed in 99b016345a77702190f894b7c3e768254bba98ce

src/client/components/pages/Courses/ViewModal.tsx
rmainseas: Ah, I just realized what you mean about this not having the unsaved changes behavior....

I'm not sure if we talked about that aspect of things or not....personally, I'd say it was probably fine since clicking "Done" simply dismisses the modal anyway and besides - there aren't really any unsaved changes since the changes the user makes take effect immediately.

natalynyu: Ok, it seems like this makes the most sense for now, especially if "done" is just dismissing the modal.

natalynyu (approved):
Approved!

PR #454, opened 2022-01-21
Improve Course Instance Table performance

### LINKED ISSUE #443: Performance analysis of models in course instance table.

jonseitz:
This PR refactors a bunch of implementation details of the `CoursesPage`/`CourseInstanceTable` to improve performance by minimizing the number full re-renders that are triggered by prop/state changes. Under the previous implementation, the entire UI would freeze for ~1.5 seconds whenever a modal opened or closed, and whenever an AppMessage was dispatched/cleared. These changes have removed the AppMessage delays entirely, as well as the modal open delays. There's still a 1-2s delay when saving a Modal (since the complete course list needs to update) but we at least have the loading spinner in place so it's at least clear that things are happening.

I strongly suggest reading through the git commit messages, as I've tried to go into detail about why the different changes were implemented and how each affects performance.

I also recorded a quick screencast to run through how I did the performance profiling, and to demonstrate the difference in rendering times between the two implementations. That's on Google Drive in [our shared App Development folder](https://drive.google.com/file/d/1GeXKr_7eeH4vlCPk5BCW7zyKj6wWo9D5/view?usp=sharing).

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #443

### DIFF
--- a/src/client/components/pages/Courses/CourseInstanceTable.tsx
+++ b/src/client/components/pages/Courses/CourseInstanceTable.tsx
@@ -63,7 +63,7 @@ interface CourseInstanceTableProps {
   /**
    * The ref value of the edit faculty absence button
    */
-  getButtonRef: (buttonId: string) => Ref<HTMLButtonElement>;
+  setButtonRef: (nodeId: string) => (node: HTMLButtonElement) => void;
 }
 
 /**
@@ -77,7 +77,7 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
   filters,
   openMeetingModal,
   openInstructorModal,
-  getButtonRef,
+  setButtonRef,
 }): ReactElement => {
   const courseColumns = tableData.filter(
     ({ columnGroup }): boolean => (
@@ -274,9 +274,7 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
             {tableData.map(
               (field: CourseInstanceListColumn): ReactElement => {
                 const { FieldContent } = field;
-                const buttonRef = getButtonRef(`${field.key}-${course.id}`);
-                if (field.viewColumn
-                === COURSE_TABLE_COLUMN.CATALOG_NUMBER) {
+                if (field.viewColumn === COURSE_TABLE_COLUMN.CATALOG_NUMBER) {
                   return (
                     <TableRowHeadingCell
                       scope="row"
@@ -311,7 +309,7 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
                             ? openInstructorModal
                             : null
                         }
-                        buttonRef={buttonRef}
+                        buttonRef={setButtonRef(`${field.key}-${course.id}`)}
                       />
                     </CellLayout>
                   </TableCell>
--- a/src/client/components/pages/Courses/CourseInstanceTable.tsx
+++ b/src/client/components/pages/Courses/CourseInstanceTable.tsx
@@ -63,11 +63,7 @@ interface CourseInstanceTableProps {
   /**
    * The ref value of the edit faculty absence button
    */
-  buttonRef: Ref<HTMLButtonElement>;
-  /**
-   * The id of the edit button corresponding to the opened modal
-   */
-  modalButtonId: string;
+  getButtonRef: (buttonId: string) => Ref<HTMLButtonElement>;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


src/client/components/pages/Courses/CourseInstanceTable.tsx
```diff
@@ -323,4 +323,16 @@ const CourseInstanceTable: FunctionComponent<CourseInstanceTableProps> = ({
...
+ * (returning true from the second argument skips updating).
+ */
+export default React.memo(
+  CourseInstanceTable,
+  (prev, next) => (
+    prev.tableData === next.tableData
+  && prev.filters === next.filters
+  && prev.courseList === next.courseList
+  )
+);
```
natalynyu, line 338: I was seeing performance issues on [my branch](https://github.com/seas-computing/course-planner/tree/feature/424-course-text-filters), which adds text filters to the `CourseInstanceTable.` When I typed in the filter field, this updated the `filters` state, causing the entire table needing to be re-rendered in between key strokes. This caused a very noticeable delay between typing and the filter field displaying what was typed. As a workaround, I separated the Course Instance Body from the rest of the table and memoized the `CourseInstanceTableBody` instead of the entire table (commit 959362fc39e76bb927c668b58a9be3a084ed558b) so that typing in the text filter field lets the user see immediately what was typed. Maybe we can take a look at this again together next week.

natalynyu (changes requested):
This is looking great! Thank you for the screen share and the detailed commit notes - I learned a lot. I started using the performance tools myself to uncover why some delays are happening in my text filters ticket. Since I am also working on performance improvement for the text filters in the Course Instance table, it might be helpful to go over those changes together sometime next week. I think the big major difference is that I separated the Course Instance table body from the rest of the table (and memoized the `CourseInstanceTableBody` in commit 959362fc39e76bb927c668b58a9be3a084ed558b), because each keystroke in the text filter field was triggering a re-render of the entire table. This was making it so that I could not see what I typed in the text filter field until a few seconds later. I haven't found a way to fix this issue without separating the table body from the rest of the table, but perhaps there is a better solution.

src/client/components/pages/Courses/__tests__/tableFields.test.tsx
```diff
@@ -238,13 +268,14 @@ describe('tableFields', function () {
       });
       context('When semester does not have data', function () {
         it('Should return an empty list', function () {
-          const springTimes = formatMeetings(TERM.SPRING);
+          const SpringTimes = formatMeetings(TERM.SPRING);
```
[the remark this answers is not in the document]
jonseitz: Yep, that's a good idea. Changed that in 694e6e79

src/client/components/pages/Courses/CourseInstanceTable.tsx
jonseitz: I think that's a good approach. In fact, I think it might be worth going a step farther and also splitting out the `TableHead` section, which would let us be more specific about which props are passed into the separate components and which are memoized (i.e., we don't need to update the table head unless the filters/columns have changed). 

I agree that we should look at this together though, so we can try to sort out how to actually make those changes without getting into a huge merge conflict mess.

src/client/components/pages/Courses/CourseInstanceTable.tsx
```diff
@@ -63,11 +63,7 @@ interface CourseInstanceTableProps {
   /**
    * The ref value of the edit faculty absence button
    */
-  buttonRef: Ref<HTMLButtonElement>;
```
natalynyu, line 66: Since this was removed, we could remove the `Ref` import.

src/client/components/pages/Courses/tableFields.tsx
```diff
@@ -248,19 +249,23 @@ export const formatMeetings = (
...
         <FontAwesomeIcon icon={faEdit} />
       </BorderlessButton>
     </>
   );
-};
+});
 
 /**
- * Descibes the additional options passed into the getValue function
+ * Descibes the additional options passed into the FieldContent functional
```
natalynyu, line 261: Just a typo - Describes

src/client/components/pages/Courses/CourseInstanceTable.tsx
natalynyu: Sounds good! We can work through this tomorrow.

natalynyu (approved):
Everything looks good to me!

natalynyu (approved):
This looks great! Approved

PR #460, opened 2022-02-10
Add Show Retired Checkbox Filter

### LINKED ISSUE #450: Course Table: Show retired courses
This ticket implements the functionality to allow the user to show or hide retired courses in the course instance table interface.

As illustrated in the mock up (https://seas-computing.github.io/course-planner_ux/courses.html), the filter shoud be a checkbox above the table.

When the check box is NOT CHECKED, the retired courses are hidden.

When the check box IS CHECKED, the retired courses are shown in table below.

The default will be for the box to be UNCHECKED.

There is already a hard coded value setting the retired courses to be hidden (`showRetired`). This ticket will involve moving value to a controllable state object.

natalynyu:
This PR adds a checkbox filter for the `retired` field and replaces the hard coded `showRetired` value with a state value that changes as you interact with the checkbox. By default, the field is unchecked, meaning that retired courses are not shown initially.

I also added a `MenuFlex` component (not sure if the naming of this component is the best) used to add space between each of the options above the Course Instance table. As more items are added within `MenuFlex`, this component should continue to space each of those items evenly. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #450

### DIFF
--- a/src/client/components/general/flex/MenuFlex.tsx
+++ b/src/client/components/general/flex/MenuFlex.tsx
@@ -1,0 +1,13 @@
+import styled from 'styled-components';
+
+/**
+ * A wrapper around the elements of options above the Courses table , used for
+ * adding secondary layers of flex positioning.
+ */
+export default styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: baseline;
+  align-self: center;
+`;
--- a/src/client/components/general/flex/MenuFlex.tsx
+++ b/src/client/components/general/flex/MenuFlex.tsx
@@ -1,8 +1,8 @@
 import styled from 'styled-components';
 
 /**
- * A wrapper around the elements of options above the Courses table , used for
- * adding secondary layers of flex positioning.
+ * A wrapper used to add a secondary layer of flex positioning for a group
+ * of options.
  */
 export default styled.div`
   display: flex;
--- a/src/client/components/pages/Courses/CoursesPage.tsx
+++ b/src/client/components/pages/Courses/CoursesPage.tsx
@@ -442,15 +442,6 @@ const CoursesPage: FunctionComponent = (): ReactElement => {
             {' '}
             Customize View
           </Button>
-          <Checkbox
-            id="showRetiredCheckbox"
-            name="showRetiredCheckbox"
-            label="Show Retired"
-            checked={showRetired}
-            onChange={() => setShowRetired(!showRetired)}
-            labelPosition={POSITION.RIGHT}
-            hideError
-          />
           <Dropdown
             id="academic-year-selector"
             name="academic-year-selector"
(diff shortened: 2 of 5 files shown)


src/client/components/pages/Courses/CoursesPage.tsx
```diff
@@ -408,19 +411,30 @@ const CoursesPage: FunctionComponent = (): ReactElement => {
...
+                  {' '}
+                  Customize View
+                </Button>
+                <Checkbox
+                  id="showRetiredCheckbox"
+                  name="showRetiredCheckbox"
+                  label="Show Retired"
+                  checked={showRetired}
+                  onChange={() => setShowRetired(!showRetired)}
+                  labelPosition={POSITION.LEFT}
```
jonseitz, line 434: Minor point, but in other places where we've used checkboxes in the app we've used `POSITION.RIGHT`. Can we change that here so we're consistent?

src/client/components/general/flex/MenuFlex.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+
+/**
+ * A wrapper around the elements of options above the Courses table , used for
+ * adding secondary layers of flex positioning.
+ */
```
jonseitz, line 6: I think `MenuFlex` is fine as a name here, but I do think we'll probably also use this for the Schedule page so I don't think we need to specify that it's for the Courses Page in the comment.

jonseitz (changes requested):
Two very minor notes here, and it looks like there's some merge conflicts with my work on the academic year navigation stuff that need to be addressed. 

Otherwise this all looks great!

src/client/components/general/flex/MenuFlex.tsx
natalynyu: Ok that makes sense. I updated the comment in [042ce39](https://github.com/seas-computing/course-planner/pull/460/commits/042ce397ad4ba352ec39f907bd00e3573e10c796).

src/client/components/pages/Courses/CoursesPage.tsx
```diff
@@ -425,34 +428,45 @@ const CoursesPage: FunctionComponent = (): ReactElement => {
...
+            onChange={
+              ({
+                target: { value },
+              }: ChangeEvent<HTMLSelectElement>) => {
+                setSelectedAcademicYear(parseInt(value, 10));
+              }
             }
-          }
-        />
+          />
```
jonseitz, line 468: Can we just swap the order of the Academic Year Selector and the Show Retired button? I think it makes a little more sense for that to be more prominent in the middle of the page

jonseitz (changes requested):
One more small tweak on the order of components in the `MenuFlex`

src/client/components/pages/Courses/CoursesPage.tsx
natalynyu: Thanks that makes sense! I updated this in [38b5501](https://github.com/seas-computing/course-planner/pull/460/commits/38b5501c44260129d6f93cd4389ed1946995f224).

jonseitz (approved):
Looks good!

PR #472, opened 2022-03-02
Feature/470 public client

### LINKED ISSUE #470: setup of info.seas
This ticket provides the setup for the info.seas public view of the app.

In this view, only the 4 year plan and schedule are viewable.

When a user arrives at the app via `info.seas.harvard.edu` they will not be taken to the HarvardKey login. Instead they will be taken directly to the app with only the 4 year plan and schedule tabs visible.

For this ticket, an `PUBLIC_CLIENT_URL` environment variable will need to be created that will allow the app to direct requests to the appropriate authorization strategy, either login or not.

When users arrive from `info.seas.harvard.edu`, replace `Logout` button in the top right with a `Login` button. Clicking this button will redirect the user to the private version of the app, and the HarvardKey login.

There will also need to be some CORS changes, allowing the APIs to be called from the `PUBLIC_CLIENT_URL` host name.

jonseitz:
This makes changes to the `AppHeader` and `AppRouter` components to provide the world-readable version of the app when users visit the public URL. This does require adding a new environment variable, PUBLIC_CLIENT_URL, so I've marked this as a breaking change. 

For local testing in the browser, I recommend setting the following in `.env`: 

```
CLIENT_URL=http://planning.seas.harvard.edu:3000
PUBLIC_CLIENT_URL=http://info.seas.harvard.edu:3000
```

And also adding the following line to `/etc/hosts` (not sure if this file exists by default on mac, but you'll need to create it if not)

```
127.0.0.1 info.seas.harvard.edu planning.seas.harvard.edu
```

This should route all of your requests to those domains back to your localhost, so you'll be able to visit info.seas.harvard.edu:3000 in the browser to see the public-facing version of the site, and then planning.seas.harvard.edu:3000 to see the private one.

You'll probably want to remove the `/etc/hosts` line when you're done (or comment it out with a leading `#`), in case you need to access the actual versions of those sites.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #470

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -41,6 +41,10 @@ SESSION_SECRET=
 # e.g. http://localhost:3000
 CLIENT_URL=
 
+# The hostname for the world-readable version of the application
+# e.g. PUBLIC_CLIENT_URL=info.seas.harvard.edu
+PUBLIC_CLIENT_URL=
+
 # Main backend data store settings
 #   The main backend data store is used as the main data storage recepticle for
 #   this application
--- a/.mochainit.ts
+++ b/.mochainit.ts
@@ -3,7 +3,10 @@ import MockDB from './tests/mocks/database/MockDB';
 
 before(function (done) {
   process.env.NODE_ENV = 'testing';
-  process.env.SERVER_URL = '';
+  process.env.SERVER_URL = 'https://computing-apps.seas.harvard.edu/course-planner';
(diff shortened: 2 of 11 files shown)


natalynyu (approved):
Approved!

PR #473, opened 2022-03-03
Offered Modal (Excludes "Retired" Option)

### LINKED ISSUE #457: Course Instance Table: "Is Offered" edit modal
This ticket provides the functionality for a user to be able to change the "Is Offered" status of a course instance from within the course instance table. 

For this ticket, we will implement three of the four desired options:

* Blank: It is not offered in the semester and it wouldn't usually be offered in the semester
* Yes: It is being offered in the semester
* No: It is usually offered in this semester, but is not being offered in this semester in particular 

The fourth option, "Retired", will be implemented by #458.

This ticket will also implement a generic back-end API route for editing a course instance.
The API endpoint will expect the full Course Instance object and will update as necessary. We can reuse this for "Enrollment" in the future.

natalynyu:
This PR adds the offered modal into the Course Instance table along with the controller, service, and API to actually update the offered value in the database. I decided to use the `CourseInstanceUpdateDTO` as the response DTO instead of the entire `CourseInstanceResponseDTO`, because I found that I did not need the entire `CourseInstanceResponseDTO` to easily update the local course instance data for the Course Instance table.

## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #457


src/client/components/pages/Courses/OfferedModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+            >
+              <Dropdown
+                id="offered"
+                name="offered"
+                label="Edit Offered Value Dropdown"
+                options={
+                  Object.values(OFFERED)
+                    .filter((value) => value !== OFFERED.RETIRED)
+                    .map((offeredValue): {
+                      value: string; label: string
```
jonseitz, line 190: Technically we don't need the return type here because it can be inferred by the immediate return value, but since we have it we should have `value: OFFERED`.

src/common/dto/courses/CourseInstanceUpdate.dto.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+export default abstract class CourseInstanceUpdateDTO {
+  @ApiProperty({
+    type: 'string',
+    example: OFFERED.Y,
+    enum: OFFERED,
+  })
+  @IsEnum(OFFERED)
+  @IsNotEmpty()
```
jonseitz, line 21: This is actually incompatible with `@IsEnum(OFFERED)`, since one of the valid values is `OFFERED.BLANK`, which is a blank string.

So at present, trying to set a course instance's `offered` field to `OFFERED.BLANK` is returning a 400 validation error, which, incidentally, is breaking the app. I think it's an issue with how validation errors are being handled in the modal component itself...

src/server/courseInstance/__tests__/courseInstance.service.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      const result = await courseInstanceService.editCourseInstance(
+        cs50CourseInstance.id, {
+          ...cs50FallInstanceUpdate,
+          offered: newOfferedValue,
+        }
+      );
+      deepStrictEqual(result, expectedResponse);
+    });
+  });
+});
```
jonseitz, line 93: As noted in the DTO, trying to update an instance's offered value to `OFFERED.BLANK` is returning a validation error, and I think that's one place where only having unit tests of the controllers/services can lead to a lot of holes in our test coverage (not the test coverage number, specifically, but just the general confidence we should have that our tests are catching errors).

I'd suggest that, instead, we probably want to use something like in `tests/server/meetings/meeting.controller.test.ts`, where we stand up the whole app and use the exposed API to send different types of test data (i.e. all of the different `OFFERED` values), then confirm that 1) the correct data is returned, and 2) the correct data is persisted to the database.

src/client/components/pages/Courses/OfferedModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+          studyCardEnrollment: currentCourseInstance[currentTermKey]
+            .studyCardEnrollment,
+          actualEnrollment: currentCourseInstance[currentTermKey]
+            .actualEnrollment,
+        }
+      );
+    } catch (error) {
+      if (axios.isAxiosError(error)) {
+        const serverError = error.response.data as Error;
+        setErrorMessage(serverError.message);
```
jonseitz, line 144: It looks like in the case of a validation error, the value of `serverError.message` is an array of Objects, each corresponding to a different validation issue. Since we're dropping this straight in to the `ModalMessage` below, it's resulting in a White Screen of Death since arrays/objects can't be used directly as children. So we probably need some additional parsing here to handle validation errors specifically.

(This is set up the same in our other modals, so this has probably been a sleeping bug for a while that's only popping up now -- I wouldn't worry about updating those other instances in this PR, but we probably want to open a ticket for that...)

jonseitz (changes requested):
A couple of typing notes below, and a surprise bug in how we're handling validation errors that needs to be addressed here and, eventually in our other modals.

And there isn't anywhere to note this inline, but we'll want to add the new `PUT` route to the `endpointTests` array in `tests/integration/server/auth/auth.test.ts`, with something like:

```
  {
    path: `/api/course-instances/${dummy.uuid}`,
    method: 'put',
    expectAnonymousAccess: false,
    expectAllowedGroup: GROUP.ADMIN,
  },
```

src/client/components/pages/Courses/OfferedModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+    offered: '',
+  });
+
+  const updateFormFields = (event: ChangeEvent): void => {
+    const target = event.target as FormField;
+    setFormFields({
+      ...form,
+      [target.name]: target.value,
+    });
+  };
```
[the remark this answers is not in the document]
natalynyu: Thanks for all your notes on types and type assertions. Those are really great points that I hadn't considered before. I removed the casting in [f54227d](https://github.com/seas-computing/course-planner/pull/473/commits/f54227d71632bcb87cc904d18efdff90cde44829) and fixed the type in the initial `useState` for offered statement in [f0cc07e](https://github.com/seas-computing/course-planner/pull/473/commits/f0cc07ef0f9f2c7638d848135bb3485b7f72c8fa).

src/server/courseInstance/__tests__/courseInstance.service.test.ts
natalynyu: Thanks for catching this! I added tests for all the different options through [60dc4a3](https://github.com/seas-computing/course-planner/pull/473/commits/60dc4a325b8119b5f3d115cc264c94b986f9e166) so that we we checking each of the possible values for `OFFERED`.

jonseitz (approved):
This all looks good. Approved!

PR #485, opened 2022-03-29
Enable SSL support for production deployment

### LINKED ISSUE #482: Listen on HTTPS from node container
Per a discussion with the Ops Team over Slack, we need the traffic between our application container and our load balancer to be encrypted in Production. [Nest is capable of listening over https][nesthttps], but we'll need to set up additional environment variables for the certificate and key.

[nesthttps]: https://docs.nestjs.com/faq/multiple-servers

jonseitz:
This PR rolls up a couple different changes that are needed to get the app up and running in production correctly, some of which came out of a working session I had with John Fisher (@jfisherseas). I considered breaking this into 3-4 separate PRs, but the two major ones are similar enough that I thought they could fit as one, and the other bits aren't pretty minor.

**Most importantly, these changes will require destroying and rebuilding the development database locally.**

## Serving the app over HTTPS

Our server container is going to run behind a load balancer in production, and that load balancer will handle incoming HTTPS connections from clients with a valid certificate. We also want to encrypt the traffic from the container to the load balancer, but we don't actually care whether the certificate between those two is valid.

To turn on HTTPS, we need to set `HTTPS_ON` to `"true"`, then provide the certificate and key as `HTTPS_PUBLIC_CERT` and `HTTPS_PRIVATE_KEY`, respectively. In production, those will come from AWS Secret Manager. To generate those values locally, you can run*:

```sh
# from https://wiki.archlinux.org/title/OpenSSL#Generate_a_self-signed_certificate_with_private_key_in_a_single_command
$ openssl req -x509 -newkey rsa:4096 -days 3650 -subj "/CN=localhost" -nodes -keyout private.key -out public.crt
```

and then copy the contents into the correct variables.

> \* I'm not actually sure if macOS has openssl installed by default, or if the option flags are the same. If that doesn't work, let me know and we can troubleshoot

You'll also need to change the `SERVER_URL` variable to use `https://`, so that the client will try to connect to the correct address. The browser may complain about the self-signed certificate, in which case you can try going directly to `{SERVER_URL}/health-check` and add an exception for the certificate (usually requires clicking on "Advanced", then "proceed anyway", or something like that).

## Connecting to the database over SSL

On a similar tack, we want to encrypt the connection from our container to the database and we don't actually care much about whether the certificate we use for that is valid. From the node side, we simply have to provide a value to the `ssl` key in the TypeORM connection options; in this case an object that sets `rejectUnauthorized: false`, which just means that we're OK with a self-signed certificate.

I went down the road of conditionally un-setting this in development and testing, but ultimately decided it would probably be better to just turn on SSL in all of our different datbases, so that our testing environment will more closely mirror production.

To that end, I've added the `init-postgres-ssl.sh` script, which is essentially running the same command as above to generate the certificate and key in the files that postgres expects. That gets mounted in the `initdb` directory on the container, and will be run automatically, **but only if container starts with an empty database**. So to enable it, you'll need to run:

```sh
$ docker-compose down postgres
$ docker volume rm course-planner_postgres_data
$ docker-compose up postgres
```

Then rerun the migrations and the `course-planner-etl`.

## Minor bugs

Outside those SSL changes, there was an errant import of `testData` in one of the DTOs that was breaking the app, a small change in the global prefix, and a change to the logger settings to get proper diagnostic information when things break.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #482

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -9,6 +9,24 @@ CAS_URL=
 # The environment the app is running in. Can be "development" or "production"
 NODE_ENV=
 
+# Whether the app should listen via https on the SERVER_PORT. Should be "true"
+# to listen for https, and "false" for plain http. Leaving it blank, or using
+# any other value, will be be the same as "false". When set to true, you will
+# also need to provide the HTTPS_PUBLIC_CERT and HTTPS_PRIVATE_KEY values below
+HTTPS_ON=false
+
+# The Public certificate that will be used to serve the app over https when
+# HTTPS_ON is true. This can be left blank in development. Should be a string in
+# PEM format. See `cert` argument to tls.createSecureContext()
+# https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions
+HTTPS_PUBLIC_CERT=
+
+# The Private key that will be used alongside the public cert when HTTPS_ON is
+# true. Can be left blank in development. Should be a string in PEM format. See
+# `key` argument to tls.createSecureContext()
+# https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions
+HTTPS_PRIVATE_KEY=
+
 # The maximum level of log that should be shown. Must be one of the following
 # strings:
 #
--- a/.env-example
+++ b/.env-example
@@ -9,6 +9,22 @@ CAS_URL=
 # The environment the app is running in. Can be "development" or "production"
 NODE_ENV=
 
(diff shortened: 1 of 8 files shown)


natalynyu (approved):
I was able to follow each of the steps listed - thank you for the detailed commits and links to documentation. Everything looks good to me!

PR #486, opened 2022-04-08
Allow users to add notes to a course

### LINKED ISSUE #453: Course instance table: Allow users to add notes to a Course
This ticket provides the ability for a user to add notes to a Course (as opposed to a Course Instance). 

This should utilize the Mark One modal and text area components.

Styling should be as demonstrated in the prototype: https://seas-computing.github.io/course-planner_ux/courses.html

The modal should include two buttons: "Cancel", and "Save"

## Cancel
* Clicking cancel, or clicking outside of the modal will trigger the unsaved changes notifications.
* Confirming unsaved changes will close the modal without any changes to the data.

## Save
* Clicking "Save" will send the value of the text area field to the `Course` PUT request endpoint.
* This ticket will require updating the `Course` PUT request endpoint to include the `notes` field.

### Validation

* On submit of the notes data, we should trim blank spaces.

rmainseas:
This PR adds a notes modal to the course instance table to allow users to add notes to courses.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #453


src/client/components/pages/Courses/NotesModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+            });
+            onSave({
+              ...course,
+              notes: courseNotes,
+            });
+          }}
+          variant={VARIANT.POSITIVE}
+        >
+          Save
+        </Button>
```
natalynyu, line 123: I think it would be nice to have the loading spinner showing as the notes are saving. (This was implemented in other course modals)

natalynyu (changes requested):
This looks and functions great!

I checked the ticket spec, and it includes adding the unsaved changes warning, so it looks like this is something that we would like to add. I think it would also be nice to display the loading spinner when a user clicks the save button.

src/client/components/pages/Courses/CoursesPage.tsx
```diff
@@ -227,6 +239,25 @@ const CoursesPage: FunctionComponent = (): ReactElement => {
...
+
+  /**
+   * Handles closing the notes modal and setting the focus back to the button
+   * that opened the modal.
+   */
+  const closeNotesModal = useCallback(() => {
+    setNotesModalData({ visible: false });
+    setTimeout(() => {
+      document.getElementById(modalButtonId)?.focus();
+    });
```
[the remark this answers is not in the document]
rmainseas: So, the reason I used `document.getElementById` is that the `closeMeetingModal` and `closeInstructorModal` callbacks use the following code:

```js
if (modalButtonId && modalButtonId in refTable.current) {
  refTable.current[modalButtonId].focus();
}
```

This looks up `modalButtonId` in `refTable.current. I did try doing this before, but it didn't work because the button ID is never stored in `refTable.current` to begin with. As best I can determine, this value is initially populated by [`setButtonRef`](https://github.com/seas-computing/course-planner/blob/develop/src/client/components/pages/Courses/CoursesPage.tsx#L172-L178) which is passed to [`CourseInstanceTableBody`](https://github.com/seas-computing/course-planner/blob/develop/src/client/components/pages/Courses/CourseInstanceTableBody.tsx) and called [here](https://github.com/seas-computing/course-planner/blob/develop/src/client/components/pages/Courses/CourseInstanceTableBody.tsx#L108). Since the notes modal is not part of `CourseInstanceTableBody` this is never called.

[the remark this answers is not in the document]
rmainseas: Looks like it's working now! Fixed in e7fe18c8172ac68ac87872a1e226dd359f00b226 and edf0bee252b1466117bb289d12803e3f84757d4d

src/client/components/pages/Courses/NotesModal.tsx
rmainseas: Great idea, added in 26fe2dc77f013bf259d93e9dc2beb254912d9151

natalynyu (changes requested):
The unsaved changes and loading spinner are working, but it seems there is something going on with the saving. When I try to save some notes, I get a bad request error:
image
Are you getting this same error?

rmainseas:
> The unsaved changes and loading spinner are working, but it seems there is something going on with the saving. When I try to save some notes, I get a bad request error: image Are you getting this same error?

image


I'm not, no. But I _am_ seeing `undefined undefined` in the catalog number, which probably needs to be addressed...

natalynyu (approved):
Everything looks good to me. Approved!

src/client/components/pages/Courses/NotesModal.tsx
```diff
@@ -166,42 +172,52 @@ const NotesModal: FunctionComponent<NotesModalProps> = function ({
...
+                isDisabled={!isEditable}
                 onChange={
-                  ({ target: { value } }: ChangeEvent<HTMLTextAreaElement>) => {
-                    setIsChanged(true);
-                    setCourseNotes(value);
+                  ({
+                    target: { value, attributes },
+                  }: ChangeEvent<HTMLTextAreaElement>) => {
+                    const disabled = !!attributes.getNamedItem('disabled');
+                    if (!disabled) {
```
jonseitz, line 181: Instead of the quadruple-negation here, what about using the `isEditable` value directly, e.g.

```suggestion
                    if (isEditable) {
```

That seems like it will be easier to parse to future maintenance, and ensures that we're always using the same value (i.e. the `isEditable` prop) to control each part of the modal

src/client/components/pages/Courses/tableFields.tsx
```diff
@@ -612,7 +612,7 @@ export const tableFields: CourseInstanceListColumn[] = [
     }: FieldContentProps): ReactElement => {
       const { notes } = course;
       const hasNotes = notes && notes.trim().length > 0;
-      const titleText = hasNotes ? 'View/Edit Notes' : 'Add Notes';
+      const titleText = `${!hasNotes ? 'Add' : 'View/Edit'} notes for ${course.catalogNumber}`;
```
jonseitz, line 615: Now that we're limiting who can actually edit notes, the "Add" and "Edit" text doesn't quite feel right. What if we just simplify it something like "Open Notes for ${course.catalogNumber}"? Otherwise trying to cover all the conditional cases feels like it would get messy very quickly...

jonseitz (changes requested):
Just a couple small notes aimed at streamlining a few of the conditional statements

src/client/components/pages/Courses/NotesModal.tsx
rmainseas: I did think about doing this but my decision not to was based on attaching the blocking of the handler to the `disabled` prop. That way the handler fires or doesn't fire based on the presence of the field being `disabled` and it behaves more like you'd expect.

I alluded to this in the commit message of af7213bcdbb653d95fa63ad4549c261a67317f9d but I don't think I outlined that specifically.

tests/integration/client/courseInstance/notesModal.test.tsx
```diff
@@ -25,89 +30,158 @@ describe('Notes modal', function () {
...
     getStub = stub(CourseAPI, 'getCourseInstancesForYear');
     postStub = stub(CourseAPI, 'editCourse');
     getStub.resolves([testData]);
     postStub.resolves();
+  });
+  describe('open/close behaviour', function () {
+    beforeEach(async function () {
+      page = render(<CoursesPage />);
+      editNotesButton = await page.findByLabelText(
+        /(Add|View\/Edit) notes for.*/
```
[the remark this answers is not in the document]
rmainseas: Oh, I didn't know that. Fixed in 9ad9046be429b71c6cbf29f684b8e93cda3fc9a5

src/client/components/pages/Courses/tableFields.tsx
rmainseas: Good point. Fixed in bde7402350b4de17a293a43a4258889221a85763

src/client/components/pages/Courses/NotesModal.tsx
rmainseas: @jonseitz Addressed this in 515c415

jonseitz (approved):
LGTM

PR #489, opened 2022-04-13
Feature/build client bundle

jonseitz:
As discussed during an Ops/Dev meeting, we're going to bundle our client-side code with a github action similar to what's being done in the [sec-directory-client][secdc] project. The minor hitch here is that we actually need to bundle two different versions of the code, one with the URLs for prod baked in and one with URLs for QA baked in. Those URLs are being set as repository secrets -- not because they're _secret_, necessarily, but just because that's the only way to provide them in a more parameterized fashion so that we can reuse this workflow later in other projects.

On top of this change, I realized that the docker image tagging was implemented incorrectly. It was always using the `main` branch but always applying the `stable`, `qa`, and `semver` tags to that image; which breaks our whole versioned deployment concept. Now there's a separate 

Unfortunately, testing for a change like this is tricky, as workflows that use the `workflow_run` trigger need to be committed to the `main` branch before they'll execute. I [forked the repo][fork] and merged these changes to `main`, then made some trivial changes to the code and ran through the `feature => develop => main` flow to make sure that the correct container image and tarball were being published at each stage. You can repeat that experiment, or just look through the [Actions tab][actions], the [releases page][releases], and the [container repository][ghcr] to see that outputs match expectation.

[secdc]: https://github.com/seas-computing/sec-directory-client/blob/main/.github/workflows/test-build.yml#L26-L90
[fork]: https://github.com/jonseitz/course-planner-action-test
[actions]: https://github.com/jonseitz/course-planner-action-test/actions
[releases]: https://github.com/jonseitz/course-planner-action-test/releases
[ghcr]: https://github.com/jonseitz/course-planner-action-test/pkgs/container/course-planner-action-test

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #___

### DIFF
--- a/.github/workflows/04-documentation.yml
+++ b/.github/workflows/04-documentation.yml
@@ -11,6 +11,7 @@ jobs:
   publish:
     name: Generate Documentation
     runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'success'
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/05-docker.yml
+++ b/.github/workflows/05-docker.yml
@@ -9,25 +9,40 @@ on:
     types:
       - completed
 jobs:
-  publish:
+  # Determines current version, and whether it should be a QA or Production release
+  pre-publish:
     runs-on: ubuntu-latest
     if: github.event.workflow_run.conclusion == 'success'
+    outputs:
+      PACKAGE_NAME: ${{steps.PACKAGE_NAME.outputs.result}}
+      VERSION_TAG: ${{steps.VERSION_TAG.outputs.result}}
+      DOCKER_TAG: ${{steps.DOCKER_TAG.outputs.result}}
     steps:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: Set package name
+        id: PACKAGE_NAME
+        run: |
+          export PACKAGE_NAME=`echo ${{ env.GITHUB_REPOSITORY }} | gawk -F/ '{print $2}'`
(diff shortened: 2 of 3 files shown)


rmainseas (commented):
As you aluded to in your PR description, this is a bit tricky to test, but I did fork the repo, merge the changes and watch what happens as you suggested and everything looks great. The only thing I did run into was an error about username and password from the publish docker image step. 

image

I assume I'm missing some credentials somewhere. I went poking about in the official repo but couldn't find anything related to GHCR credentials. We can catch-up about this when you're back.

jonseitz:
Ah, I see what happened. That's actually a failure on [the _old_ docker publish workflow][old1], since that particular stage always uses the version on `main` and the changed had only been merged into `develop` at that point. That old version did rely on a separate personal access token (called [`CR_PAT` on line 39][L39]). At the time, that was the only way to authenticate with the then-new GitHub Container Repository, but that's since changed and the default `GITHUB_TOKEN` used by all actions is able to publish now. 

I should have noted that change somewhere, but it wound up getting lumped in with all the others. Sorry, that's my bad. 

That being said, it looks like [the _new_ workflow][new1] ran fine, so it does make things a bit more robust for the future.

[old1]: https://github.com/rmainseas/course-planner-actions-test/actions/runs/2169233113/workflow
[L39]: https://github.com/rmainseas/course-planner-actions-test/actions/runs/2169233113/workflow#L39
[new1]: https://github.com/rmainseas/course-planner-actions-test/actions/runs/2169275479

rmainseas (approved):
This looks great to me. Merge away.

PR #497, opened 2022-05-03
Enable Redis over TLS

jonseitz:
For background: When we were trying to run the course planner on ECS, we kept seeing gateway timeout errors from the container. We ultimately traced these down to the connection to redis, which was configured to only accept TLS connections, but the container was running with `NODE_ENV=development`, which disables harvard key checking but also disables TLS for redis connections. 

Rather than try to create another set of flags to conditionally enable/disable redis-over-TLS, I figured it makes more sense to set up our local redis instance to also use TLS, similar to what we did with postgres. That brings our local dev environment closer in line with the final production environment, which can potentially help us catch errors like this in the future.

Actually setting up TLS was a little more complicated than I expected though:

1. The redis image needed to be updated to a `>=6.0` version, in this case `6.2`.
2. I copied the `init-postgres-ssl.sh` script over, and changed the paths to use `/redis/` values instead.
3. There is no autoloaded script in the `bitnami/redis` image that we can use to run this, so we need to override the existing `CMD` option in [the `binami/redis` Dockerfile][dockerfile] so that [`entrypoint.sh`][entrypoint.sh] will run our script first, then execute the [`setup.sh`][setup.sh] and [`run.sh`][run.sh] scripts to actually start up the redis server.
4. Even once that was setup, the node process couldn't connect because it wouldn't accept the self-signed certificate, so we needed to further change the `ConfigService` to set `tls.rejectUnauthorized: false` when creating the redis client.

After all that, the app and its redis connection should run exactly the same as it did before, though now it will be connecting over TLS to guard our precious session ids.

[dockerfile]: https://github.com/bitnami/bitnami-docker-redis/blob/master/6.2/debian-10/Dockerfile
[entrypoint.sh]: https://github.com/bitnami/bitnami-docker-redis/blob/master/6.2/debian-10/rootfs/opt/bitnami/scripts/redis/entrypoint.sh
[setup.sh]: https://github.com/bitnami/bitnami-docker-redis/blob/master/6.2/debian-10/rootfs/opt/bitnami/scripts/redis/setup.sh
[run.sh]: https://github.com/bitnami/bitnami-docker-redis/blob/master/6.2/debian-10/rootfs/opt/bitnami/scripts/redis/run.sh

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High 

## Related Issues:

Fixes #___

### DIFF
--- a/src/server/config/config.service.ts
+++ b/src/server/config/config.service.ts
@@ -198,7 +198,7 @@ class ConfigService {
   }
 
   /**
-   * Return the redis connection string
+   * Return the redis client options
    * NOTE: This is needed to properly connect to redis over TLS
    */
 


src/server/config/config.service.ts
```diff
@@ -201,21 +202,22 @@ class ConfigService {
...
     } = this.env;
-    const redis = new URL(`redis://${REDIS_HOST}:${REDIS_PORT}`);
-    if (NODE_ENV === 'production') {
-      redis.protocol = 'rediss:';
-    }
+    const redis = new URL(`rediss://${REDIS_HOST}:${REDIS_PORT}`);
     if (REDIS_PASSWORD) {
       redis.password = REDIS_PASSWORD;
     }
-    return redis.toString();
```
rmainseas, line 218: Minor point but the docblock for this getter says: _"Return the redis connection string"_. Since it's returning `ClientOpts` this might want updating..

rmainseas (changes requested):
Looks great to me, just a quick note about the docblocks and it's all good to god

src/server/config/config.service.ts
jonseitz: Still worth changing. Did that in 5d9ca3e

rmainseas (approved):
LGTM!

Issue #504, opened 2022-05-23
Schedule: Pop up room info on all instances


jonseitz:
When clicking on a course in the schedule, each recurrence of the meeting should also pop up with room info. We do not necessarily need to replicate the multi-popover feature for screen readers

jonseitz:
Will involve changing the state value used to track which session has been clicked. Will also need work on tests to handle multiple occurrences (also adding those to the test data, potentially)

Vittorio2015:
## Accessibility review of each course pop-up with room info in "[Schedule](https://planning-qa.seas.harvard.edu/courses/schedule)"

### Monday 9 am schedule

As the focus is moved on a course via the tab key, the screen reader reads "AC 209b on Monday, 9:00 AM to 10:15 AM in 1414 Massachusetts Avenue 437"

<img width="114" alt="image" src="https://user-images.githubusercontent.com/12190138/175108928-6974c104-66a0-4cb5-9c80-08056558875e.png">

PR #510, opened 2022-06-03
Add meeting info popups to Schedule

### LINKED ISSUE #503: Schedule: Pop up room info on click
Each course listed in the schedule should be clickable, and clicking the course will display the full catalog number, date/time, and meeting room using the `<Popover>` component from #495.

The content of the Popover should be accessible to screen readers, potentially as `InvisibleText` on the page before the Popover opens.

This ticket will only show the popover next to the course clicked. Additional popovers for recurrences will be added later.

jonseitz:
This puts in the place the initial Popover effect on the schedule page, displaying a single information box next to a course number when clicking on it. A future ticket (#504) will add boxes next to every meeting of the same course in the schedule.

This will still need an accessibility review, but the current implementation:

1. Hides the button containing the course number and the popover from screen readers
2. Replicates the popover text inside of a "HiddenText" block that should replace the entire content of the list item.

This should give screen readers access to the same information contained in the popover, without needing to go through the click-to-display process.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #503

### DIFF
--- a/src/client/components/pages/Schedule/ScheduleView.tsx
+++ b/src/client/components/pages/Schedule/ScheduleView.tsx
@@ -1,9 +1,12 @@
-import React, { FunctionComponent } from 'react';
+import React, { FunctionComponent, useState } from 'react';
 import { ScheduleViewResponseDTO } from 'common/dto/schedule/schedule.dto';
 import DAY, { dayEnumToString } from 'common/constants/day';
+import { Popover } from 'mark-one';
 import {
-  WeekBlock, DayBlock, CourseListing, SessionBlock,
+  WeekBlock, DayBlock, CourseListing, SessionBlock, CourseListingButton,
 } from './blocks';
+import { PGTime } from '../../../../common/utils/PGTime';
+import { HiddenText } from '../../general';
 
 interface ScheduleViewProps {
   /**
@@ -61,6 +64,12 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
   // css-grid rows
   const numRows = ((lastHour - firstHour) * 60)
       / minuteResolution;
+
+  /**
+   * Track the prefix and catalog number of the course whose details should be shown.
+   */
+  const [currentPopover, setCurrentPopover] = useState('');
+
   return (
     <WeekBlock
       firstHour={firstHour}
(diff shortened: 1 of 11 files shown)


src/client/components/pages/Schedule/ScheduleView.tsx
```diff
@@ -95,19 +109,58 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
...
+                        >
+                          {courseNumber}
+                        </CourseListingButton>
+                        <Popover
+                          aria-hidden
+                          xOffset="0.5rem"
+                          yOffset="1rem"
+                          title={catalogNumber}
+                          isVisible={currentPopover === courseId}
+                        >
```
natalynyu, line 153: Should we calculate where the user has clicked so that the `Popover` can show up on the left side for the items on the right of the schedule?

natalynyu (changes requested):
This looks really great! I love the way that the `Popover` is styled. I just had a few comments.

src/client/components/pages/Schedule/ScheduleView.tsx
```diff
@@ -95,19 +109,58 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
...
+                    const displayStartTime = PGTime.toDisplay(
+                      `${startHour}:${startMinute.toString().padStart(2, '0')}`
+                    );
+                    const displayEndTime = PGTime.toDisplay(
+                      `${endHour}:${endMinute.toString().padStart(2, '0')}`
+                    );
+                    return (
+                      <CourseListing key={courseId}>
+                        <CourseListingButton
+                          aria-hidden
```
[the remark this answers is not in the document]
jonseitz: Ah, that's a really good point. I hadn't considered how those would interact. I'll look at revising this.

src/client/components/pages/Schedule/ScheduleView.tsx
natalynyu: Ok got it, thanks for clarifying!

natalynyu (approved):
Looks good! The tests seem to have timed out, but otherwise approved!

PR #512, opened 2022-06-06
Show server errors in edit notes modal

### LINKED ISSUE #500: Bug: Show update error in Notes modal
Currently, the Notes modal does not display any errors from the server on saving. We should correct this, to handle validation errors as well as any server/auth errors.

For validation errors, we should concatenate all the field errors into a single string. Other errors should display the message.

rmainseas:
This PR Fixes an issue where a server-side error thrown by the server (be that a validation error or general 500) can break the modal update code but with no indication to the user as to the nature of the problem. The fix for this is to concat all the validation errors into a single string and show them beneath the modal or(in the case of a general 500 error) just show a generic error message below the text area.

Since this code is ~~shamelessly ripped from~~ very similar to [the code for showing server-side errors in the enrollment modal](https://github.com/seas-computing/course-planner/blob/develop/src/client/components/pages/Courses/EnrollmentModal.tsx#L274-L293) I do just want to call out one key difference which is that unlike the enrollment modal, no effort is being made here to convert field names from camelCase to something more human readable. This is because the camelCase conversion in the enrollment modal operates on fields already known to the component, whereas in this instance, the errors tend to be from fields other than the ones controlled by this modal. All fields _could_ technically be brought into the modal and sanitized, but that would add additional complexity and logic to what should be a fairly simple component.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #500

### DIFF
--- a/src/client/components/pages/Courses/NotesModal.tsx
+++ b/src/client/components/pages/Courses/NotesModal.tsx
@@ -229,17 +229,22 @@ const NotesModal: FunctionComponent<NotesModalProps> = function ({
                 } catch (error) {
                   if ((error as AxiosError).response) {
                     const { response } = error as AxiosError;
-                    const errors = [
-                      ...(response.data as BadRequestInfo).message,
-                    ].map(
-                      ({ constraints }) => Object.entries(constraints)
-                        .map(([, message]) => message)
-                        /* We're not making any effort here to try and resolve
-                        * the field name to a human readable name because the
-                        * field name could be any field from the course object
-                        * rather than a known set of pre-determined fields */
-                    ).reduce((acc, val) => acc.concat(val), []);
-                    setModalErrors(errors);
+                    if (response.status === HttpStatus.BAD_REQUEST) {
+                      const errors = [
+                        ...(response.data as BadRequestInfo).message,
+                      ].map(
+                        ({ constraints }) => Object.entries(constraints)
+                          .map(([, message]) => message)
+                          /* We're not making any effort here to try and resolve
+                          * the field name to a human readable name because the
+                          * field name could be any field from the course object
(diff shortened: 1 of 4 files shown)


src/client/components/pages/Courses/NotesModal.tsx
```diff
@@ -1,3 +1,5 @@
+import { HttpStatus } from '@nestjs/common';
```
jonseitz, line 1: I'd prefer not to use any of the `@nestjs` or server-specific packages in our client-side code, because: 

1. It keeps our frontend app backend-agnostic 
2. It could potentially cause problems down the line if any node-specific code gets added to the package we're importing

There is a standalone [`http-status-codes`][hsc] package we could install, or we can just define our own constant to use in place.


[hsc]: https://www.npmjs.com/package/http-status-codes

jonseitz: (I did miss this in the enrollment modal PR. We don't need to update that one now, but we probably should in a future bugfix)

jonseitz (changes requested):
Just one note about the import of a nest package.

src/client/components/pages/Courses/NotesModal.tsx
rmainseas: Yep. Totally agree with all of this. I added an enum in f6d3fbcbb74040e35a9eb503cc2d583d6b192a01 to take care of this and swapped `@nestjs/common` out for it in 0dd6110a07ffda5668996e8ddbb3cf856464e8aa

jonseitz (approved):
This looks great! Thanks for adding those enum values.

PR #513, opened 2022-06-06
Make termPattern optional

### LINKED ISSUE #499: Bug: Term pattern handling
Term pattern being null, trying to save a course notes (or any other course update) returns a validation error. We should provide a default value for the term pattern and apply that in the ETL.

rmainseas:
This PR makes the `termPattern` field of the `UpdateCourse` DTO. The requirement of this field to not be null was causing erroneous validation errors to be thrown when trying to update course instances that had a null value for this field. This null value was originating in data imported from the old course planner system indicating null values are a perfectly acceptable value for this field. The database schema also allows for this, so the API disallowing null values here is likely an oversight(especially since this value isn't actually used by the app).

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #499

### DIFF
--- a/src/common/dto/courses/UpdateCourse.dto.ts
+++ b/src/common/dto/courses/UpdateCourse.dto.ts
@@ -92,6 +92,7 @@ export abstract class UpdateCourseDTO {
     enum: TERM_PATTERN,
   })
   @IsEnum(TERM_PATTERN)
+  @IsOptional()
   public termPattern: TERM_PATTERN;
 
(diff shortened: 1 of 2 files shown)


rmainseas:
It looks like this broke a test that was designed to make sure the lack of this field threw an error.....which is curious.

jonseitz (approved):
This is all set!

PR #518, opened 2022-06-08
Import HTTP status codes from api/

### LINKED ISSUE #517: Remove dependency on @nestjs/core
### Description
Client-side dependency on `@nestjs/core` is breaking the client bundle build. The client-side code is depending on this in order to get HTTP status code numbers. Instead, we should switch this to use the enum defined in https://github.com/seas-computing/course-planner/blob/main/src/client/api/httpStatus.ts.

---

- [ ] Test
- [ ] Docs

rmainseas:
HTTP status codes were re-defined in api/ for client-side code rather than importing the enum defined in `@nestjs/core`. This is because importing `@nestjs/core` in client-side code caused build failures.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #517


rmainseas:
Thanks!

PR #520, opened 2022-06-08
Fix addAcademicYear Unhandled Promise Rejection Warning

### LINKED ISSUE #519: Bugfix: Unhandled Promise Rejection in addAcademicYear
The following error is getting thrown as a unhandled promise instead of getting logged: ```UnhandledPromiseRejectionWarning: Error: Cannot create requested academic year until preceding academic year is created."```

We would like to log the errors that occur as a result of calling `addAcademicYear,` but still have the application continue running if it fails.

natalynyu:
Originally without the `await` preceding the call of `addAcademicYear` in `onApplicationBootstrap`, the error thrown in `addAcademicYear` was not being handled properly. Adding this `await` makes it so that we can catch the error and log it appropriately.

The following error was getting thrown as a unhandled promise instead of getting logged: ```UnhandledPromiseRejectionWarning: Error: Cannot create requested academic year until preceding academic year is created."```

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High 

## Related Issues:

Fixes #519

### DIFF
--- a/src/server/semester/semester.service.ts
+++ b/src/server/semester/semester.service.ts
@@ -262,12 +262,12 @@ export class SemesterService implements OnApplicationBootstrap {
     return null;
   }
 
-  public onApplicationBootstrap(): Promise<void> {
+  public async onApplicationBootstrap(): Promise<void> {
     const today = new Date();
     if (this.config.isProduction && today.getMonth() === MONTH.JUN) {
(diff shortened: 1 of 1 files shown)


jonseitz (approved):
This is all set!

PR #522, opened 2022-06-08
Avoid concurrent workflows

### LINKED ISSUE #521: Avoid running simultaneous test workflows
There seems to be a strong correlation between multiple testflows running and the appearance of timeout errors in those tests. While there may need to be some additional changes in the actual test code to resolve an underlying issue, [the `concurrency` option](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) in Github Actions could be used to at least prevent multiple instances of the same workflow from running at the same time.

jonseitz:
The [concurrency][1] field prevents multiple instances of the same
workflow (or any two workflows with the same concurrency value) from
running at the same time. It will also cancel/replace pending workflows
when the new one attempts to run. That could put us in an edge case
where a PR test run could be skipped, but I think a skipped test run is
probably better in the long run than a failed test run. In either case,
this should not affect any active workflows.

I've added individual concurrency groups for the test, versioning,
documentation, and publication stages. The PR check workflow is simple
enough that it shouldn't have any problems running concurrently, but we
can always revisit in the future.

[1]:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #521

### DIFF
--- a/.github/workflows/01-test-suite.yml
+++ b/.github/workflows/01-test-suite.yml
@@ -8,6 +8,7 @@ jobs:
   test:
     name: Run full test suite
     runs-on: ubuntu-latest
+    concurrency: course-planner-test-suite
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
--- a/.github/workflows/03-version.yml
+++ b/.github/workflows/03-version.yml
@@ -12,6 +12,7 @@ on:
         default: 'major'
 jobs:
   update_version:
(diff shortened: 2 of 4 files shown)


rmainseas (approved):
This is good to go!

PR #527, opened 2022-06-27
Feature/504 multiple schedule popups

### LINKED ISSUE #504: Schedule: Pop up room info on all instances

jonseitz:
**Marking this as a breaking change as it requires running a new Migration**

The original implementation of the popup used the id of the Meeting entity to determine whether the popup should appear. This made the popup show over the meeting that was clicked, but not over any of the other meetings associated with the course instance. Since we want the popup to show over all of those as well, we had to switch to using the id from the Course Instance instead. That field was not part of the `ScheduleEntryView` entity, so we also need a new migration to add it. 

For testing, I also had to tweak the dummy test data to include multiple instances of at least one course.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #504


natalynyu (approved):
Everything looks great to me! Approved.

PR #531, opened 2022-07-06
Bugfix: Fix tests checking retired functionality

### LINKED ISSUE #530: Bugfix: Build is failing due to tests checking retired functionality
The [build](https://github.com/seas-computing/course-planner/runs/7201553604?check_suite_focus=true) for PR #529 is continually failing. Upon further investigation, the tests related to the setting courses to retired functionality are also failing locally.

natalynyu:
These tests were previously failing, because the tests were calculating the current academic year as 2023, while these tests are trying to retire course instances from 2022. Since we're not allowed to retire instances of the past, the current academic year has been stubbed to 2022 so that the tests no longer fail as the real academic year changes.

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High 

## Related Issues:

Fixes #530

### DIFF
--- a/tests/integration/server/courseInstance/courseInstance.controller.test.ts
+++ b/tests/integration/server/courseInstance/courseInstance.controller.test.ts
@@ -53,6 +53,7 @@ describe('CourseInstance API', function () {
   let ciRepository: Repository<CourseInstance>;
   let api: HttpServer;
   let authStub: SinonStub;
+  let configService: ConfigService;
 
   beforeEach(async function () {
     authStub = stub(TestingStrategy.prototype, 'login');
@@ -94,6 +95,7 @@ describe('CourseInstance API', function () {
     meetingRepository = testModule.get(getRepositoryToken(Meeting));
     fciRepository = testModule.get(getRepositoryToken(FacultyCourseInstance));
     ciRepository = testModule.get(getRepositoryToken(CourseInstance));
+    configService = testModule.get<ConfigService>(ConfigService);
     const nestApp = await testModule
       .createNestApplication()
       .useGlobalPipes(new BadRequestExceptionPipe())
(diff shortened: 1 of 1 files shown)


tests/integration/server/courseInstance/courseInstance.controller.test.ts
```diff
@@ -53,6 +53,7 @@ describe('CourseInstance API', function () {
   let ciRepository: Repository<CourseInstance>;
   let api: HttpServer;
   let authStub: SinonStub;
+  let configService: ConfigService;
```
farassadek, line 57: I'd like to see the constants used at the top , so I like to see `let currentAcademicYear = 2022` here. But this is not important and its up to you.

tests/integration/server/courseInstance/courseInstance.controller.test.ts
```diff
@@ -1046,18 +1048,20 @@ describe('CourseInstance API', function () {
     let newOfferedValue: OFFERED;
     let response: Response;
     let testCourseInstance: CourseInstance;
+    const currentAcademicYear = 2022;
```
farassadek, line 1051: This not that important but I prefer if you move `const currentAcademicYear = 2022;` to the top behind the `let configService: ConfigService;`. Still if you like to leave it like this then its fine.

tests/integration/server/courseInstance/courseInstance.controller.test.ts
farassadek: Also if you can put a simple comment about `currentAcademicYear`

natalynyu: Sure thing, the comment was added for `currentAcademicYear` in [3e4fe28](https://github.com/seas-computing/course-planner/pull/531/commits/3e4fe28077f0c87b76831af794fabab71be52a98) and it was moved to the top after `configService`.

tests/integration/server/courseInstance/courseInstance.controller.test.ts
natalynyu: Sure, `currentAcademicYear` was moved to be after `configService.`

PR #543, opened 2022-07-19
Faculty Page Academic Year Dropdown

### LINKED ISSUE #532: Faculty: Academic Year Dropdown
Similar to Courses tab and Schedule tab, the Faculty tab needs a way to choose between academic years. This ticket will add a dropdown to the top of the page, above the table. Defaults to current academic year (or potentially use `useStoredState` to keep last chosen year in session storage). Selecting a different academic year will re-fetch and display that year's data.

For testing, should make sure that changing the drop-down does refetch the data (i.e. calls the API with the new academic year). Look at the Courses Page tests for other relevant tests.

natalynyu:
This PR adds the academic year dropdown for the Faculty page. The logic for creating the dropdown options is the same as the academic year dropdown on the Courses page, so it was extracted to its own function and is called for both the courses and faculty academic year dropdowns. I also used the `useStoredState` hook so that the same option is selected if the user navigates away from and returns to the Faculty tab. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #532

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
         "lodash.merge": "^4.6.2",
         "lodash.set": "^4.3.2",
         "loglevel": "^1.6.8",
-        "mark-one": "^0.4.6",
+        "mark-one": "^0.4.8",
         "morgan": "^1.10.0",
         "nestjs-redis": "^1.3.3",
         "nestjs-session": "^1.0.0",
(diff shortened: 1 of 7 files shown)


jonseitz (changes requested):
This looks great! I just have one comment about using something more specific for the session storage key.

jonseitz (approved):
Looks great!

PR #544, opened 2022-07-21
Fix Schedule Display - Extend end time to 10pm

### LINKED ISSUE #535: Some courses scheduled later than 'Schedule' tab displays
Some late classes (e.g. in Fall 2020) end past the end time of the schedule display. This end time is currently hard-coded to 8pm, but we should extend it to 10pm. We'll continue to use the hard-coded value, but may explore a process to determine the latest end time in the future.

natalynyu:
This PR updates the schedule such that the end time is extended from 8pm to 10pm. According to our users, the latest end time for a course is 9:30pm. For now, we will hardcode the end time value as 10pm, but we may dynamically update the end time in the future.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #535

### DIFF
--- a/src/client/components/pages/Schedule/SchedulePage.tsx
+++ b/src/client/components/pages/Schedule/SchedulePage.tsx
@@ -21,7 +21,7 @@ import { useStoredState } from '../../../hooks/useStoredState';
  */
 
 const FIRST_HOUR = 8;
-const LAST_HOUR = 20;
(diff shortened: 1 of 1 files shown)


rmainseas (approved):
Looks good to me. Approved.

PR #546, opened 2022-07-26
Feature/533 changing no longer active

farassadek:
## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #___


src/server/faculty/faculty.service.ts
```diff
@@ -36,4 +47,69 @@ export class FacultyService {
...
+        where: { id: absenceReqInfo.id },
+      });
+
+    const updateQuery = this.absenceRepository.createQueryBuilder()
+      .update(Absence);
+
+    // Are we going to have to do NO_LONGER_ACTIVE gymnastics?
+    if (
+      [existingAbsence.type, absenceReqInfo.type]
+        .some((value) => value === ABSENCE_TYPE.NO_LONGER_ACTIVE)
```
jonseitz, line 66: Minor note: This could be `.includes(ABSENCE_TYPE.NO_LONGER_ACTIVE)`. As far as I can tell it's effectively the same, but _slightly_ easier to read while going over the code.

src/server/faculty/faculty.service.ts
```diff
@@ -36,4 +47,69 @@ export class FacultyService {
...
+        updateQuery.set({ type: ABSENCE_TYPE.NO_LONGER_ACTIVE });
+      }
+
+      // Changing FROM NO_LONGER_ACTIVE
+      if (
+        existingAbsence.type === ABSENCE_TYPE.NO_LONGER_ACTIVE
+      && absenceReqInfo.type !== ABSENCE_TYPE.NO_LONGER_ACTIVE
+      ) {
+        updateQuery.set({ type: ABSENCE_TYPE.PRESENT });
+      }
```
jonseitz, line 100: I think this case is more complicated, since it is possible to change from `NO_LONGER_ACTIVE` and a different absence type other than `PRESENT`. So if we change someone from `NO_LONGER_ACTIVE` to `PARENTAL_LEAVE` in Fall 2022, it should set Fall 2022 to `PARENTAL_LEAVE` and all future semesters to `PRESENT` --  right now it's setting all semesters to `PRESENT`. 

I think this would require two separate queries, one to update the requested semester, and another to update all future semester.

tests/integration/server/faculty/facultyapi.test.ts
```diff
@@ -0,0 +1,1062 @@
...
+            }))
+              .filter(({ semester }) => parseInt(semester.academicYear, 10)
+                < thisAcademicYear);
+
+            await request(api)
+              .put(`/api/faculty/absence/${springAbsence.id}`)
+              .send({
+                id: springAbsence.id,
+                type: ABSENCE_TYPE.TEACHING_RELIEF,
+              });
```
jonseitz, line 853: Missing a test to ensure that the requested semester actually gets updated (see note on `FacultyService`)

jonseitz (changes requested):
Nice job cleaning up the tests and getting this functionality in place -- I know the logic around updating future semesters was pretty thorny, but I think we're in a good spot there now. 

There is one issue when changing an absence from `NO_LONGER_ACTIVE` that needs to be resolved -- basically just handling the case of changing a faculty member directly to another non-`PRESENT` status.

src/server/faculty/faculty.service.ts
rmainseas: Oh, I see. My understanding from the ticket was that the current semester would be set to `ABSENCE_TYPE.PRESENT` as well as all future semesters. If that's not the case then it will indeed require two queries, yes.

src/server/faculty/faculty.service.ts
rmainseas: I think I agree, fixed in cb002aaa5124533a228f6be2d65326816f4d37cd

src/server/faculty/faculty.service.ts
rmainseas: Alright, I think I've updated this in 4a2ddc0b34a1c94f1caf9fc2f57c03801b7a7281

tests/integration/server/faculty/facultyapi.test.ts
rmainseas: Added a test for this in 4a2ddc0b34a1c94f1caf9fc2f57c03801b7a7281

jonseitz (approved):
OK, this all looks good!

PR #547, opened 2022-07-26
Make schedule blocks scrollable

### LINKED ISSUE #526: Schedule: Fix course numbers appearing outside the bounds of the block
On the [Spring 2020 view of the Schedule](https://planning-qa.seas.harvard.edu/courses/schedule), some prefix blocks are not large enough to contain all the course numbers. Should we change the font size, or implement some other fix here?

Vittorio to provide a mockup.

jonseitz:
This started as a fairly straightforward style update, but got a little more complicated thanks the weirdness of absolute and relative positioning. The final version differs in implementation from what @Vittorio2015 sketched out in #526, but achieves essentially the same effect.

I also managed to find a way to add up/down indicators to show when there's more content in either direction:

> View of Wednesday afternoon in Spring 2020

Since those indicators are defined in CSS they should affect the actual DOM output parsed by screen readers. The downside, though, is that there isn't a great way to add tests for the indicators in our current system (though I'm certainly open to suggestions).

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #526

### DIFF
--- a/src/client/components/pages/Schedule/ScheduleView.tsx
+++ b/src/client/components/pages/Schedule/ScheduleView.tsx
@@ -122,6 +122,35 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                   prefix={coursePrefix}
                   startRow={resolvedStartRow}
                   duration={resolvedDuration}
+                  popovers={
+                    courses.map(({
+                      id: meetingId,
+                      instanceId,
+                      courseNumber,
+                      room,
+                    }, listIndex) => {
+                      const displayStartTime = PGTime.toDisplay(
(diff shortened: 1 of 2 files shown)


natalynyu (approved):
Everything looks good to me! The only thing I would suggest as a potential improvement would be to restrict scrolling when the popup is up, but since the underlying course number in the schedule block is highlighted, I think this would just be a "nice-to-have" type feature.

jonseitz:
Thanks @natalynyu! I agree about the scrolling. I'll throw a ticket on the backlog to address that.

PR #542, opened 2022-07-19
Fade course blocks that don't have popups

jonseitz:
This adds `isFaded` and `isHighlighted` properties to the `SessionBlock` and `CourseListingButton` components that control the opacity setting of the element. When the user clicks on a single meeting, all other `SessionBlocks` that don't contain meetings in the same course will fade out. Within the blocks that do contain meetings in the same course, any other course numbers will be faded, while the chosen course will have a light highlight applied.

 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality
      to change)

## Checklist:

- [x] I have run `eslint` on the code
- [x] I have added JSDoc for all of my code (where applicable)
- [ ] I have added tests to cover my changes.

## Priority:
- [] Normal 
- [ ] High 

## Related Issues:

Fixes #505

PR #529, opened 2022-07-05
Add Instructional Text to Multi Year Plan tab for info.seas only

natalynyu:
This PR adds instructional text from the current course planner to the multi year plan tab when the user is visiting from info.seas. The planning.seas page will continue not displaying this instructional text.

The description from PR #472 outlines how to see the public-facing (info.seas.harvard.edu) and private-facing (planning.seas.harvard.edu) versions of course planner to see this change. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality
      to change)

## Checklist:

- [x] I have run `eslint` on the code
- [x] I have added JSDoc for all of my code (where applicable)
- [x] I have added tests to cover my changes.

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #471

PR #528, opened 2022-06-29
Feature/516 add hidden option for faculty

farassadek:
Add a show/hide retired faculty checkbox to the left side of the faculty page. `no Longer Active` Faculty on an academic year will be affected. Faculty that is not active in one semester but  active on the other will be shown  by default.  I had to add a new flex style of reverse row direction for the checkbox to be on the left side. For the test, a new faculty page test page is created that test the checkbox(given that the faculty table test is already exists).
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality
      to change)

## Checklist:

- [x] I have run `eslint` on the code
- [ ] I have added JSDoc for all of my code (where applicable)
- [x] I have added tests to cover my changes.

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:
#525 

Fixes #525

PR #548, opened 2022-07-27
Merge develop into main for version 0.8.0

### LINKED ISSUE #471: info.seas: adding instructional text to 4 year plan page

### LINKED ISSUE #504: Schedule: Pop up room info on all instances

### LINKED ISSUE #505: Schedule: Fade unpopped courses
When all instances of a course have Popovers visible, all other prefix blocks and all other courses in that same prefix block, should fade into the background.

### LINKED ISSUE #515: Bug: Times in Meeting Modal don't always update as expected
When selected a new time from the dropdown in the Meeting Modal > Meeting Times section, the start and end times do not always update to the selected values. It does not happen consistently.

Recommendations for debugging:

- adding `console.log` statements in code to print state values as they update
- Recording/profiling in React dev tools to see when state updates.

### LINKED ISSUE #516: Add "Hidden" Option for Faculty
In the current course planner, faculty can be marked as "No Longer ACtive" from the Faculty tab, which hides them from display, similar to the retired courses funcationaliy. 

In our new course planner, we have the no longer active option, but:

1) they're not being hidden from display,
2) there's no option to show them, once hidden
3) the "No Longer Active" status doesn't persist into the future.
4) Not 100% related, but we're also missing the academic year selector.

This ticket will track the meta status of the feature, which will be implemented by smaller tickets:

Request from Patrick copied below:

One additional question.  I recall from the meeting that the new tool will have the ability to un-retire a course, which is great!  Can we do the same for faculty as well?  There are a lot of non-ladder folks that are still in the tool that no longer have appointments, so it would make sense to retire them to clean up the view.  However, if they happen to come back at some point, then we can’t undo that flag, so they remain.

### LINKED ISSUE #526: Schedule: Fix course numbers appearing outside the bounds of the block

### LINKED ISSUE #530: Bugfix: Build is failing due to tests checking retired functionality

### LINKED ISSUE #532: Faculty: Academic Year Dropdown

### LINKED ISSUE #535: Some courses scheduled later than 'Schedule' tab displays

jonseitz:
## New Features
 - Show multiple popups in schedule (#527 Closes #504)
 - Add option to show "No Longer Active" Faculty (#528 Closes #516)
 - Add Instructional Text to Multi Year Plan tab for info.seas only (#529 closes #471)
 - Add Academic Year Dropdown to Faculty Page (#543 closes #532)
 - Fade out Course blocks that don't have popups (#542 closes #505)

## Bug Fixes
 - Update `mark-one` to `0.4.6` to fix meeting timeslot bug (#524 Fixes #515)
 - Fix tests checkign retired functionality (#531 fixes #530)
 - Extend Schedule Display to 10pm to fix course display (#544 fixes #535)
 - Scrolling of long course lists in schedule (#547 fixes #526)


natalynyu (approved):
Looks great! Approved.

rmainseas (approved):

PR #561, opened 2022-08-12
Omit Optional dependencies by default

### LINKED ISSUE #455: node-canvas doesn't compile on ARM (i.e. mac pro m1) so build breaks
In order to build the project on my macbook pro m1, I needed to compile node-canvas separately on my local machine following these instructions: https://github.com/Automattic/node-canvas#compiling

This then resulted in the docker build failing because by default it tries to build the container on the same architecture (arm) as the host.

In order to get the build to work I needed to change the dockerfile to build base using the specific amd64 architecture: 

`FROM amd64/node:14 as base`.

jonseitz:
This Pull Request should resolve the issues with building the course-planner container on ARM-based systems (e.g. Apple's M1 Macs). Chris provided some details on this in #455, but essentially the problem is that the `node-canvas` package being pulled in as an option dependency of `jsdom` needs to install a binary package, `cairo`, that doesn't have ARM builds available.

The current solution recommended by the project is to use the `--build-from-source` flag to `npm install`, which is _fine_ but introduces some additional overhead for a package that we're not even really using. Given that, it makes more sense to just omit the "optional" packages altogether and proceed on the assumption that any optional packages we do need should be installed explicitly.

This also lets us avoid the other fix proposed by Chris, of explicitly building on the `amd64` version of the container. From what I understand this would require an additional emulation layer when running Docker on the mac, which is already running in a virtualized environment, which could in turn cause slower performance for the container.

I tested this live with @Urjikoo (the only dev on our team with an M1) and confirmed that she is able to build the container.

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High 

## Related Issues:

Fixes #455

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+omit=optional
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,12 +10,12 @@ RUN chown node:node ${APP_DIR}
 COPY --chown=node:node package*.json ./
 RUN ["npm", "install", "--global", "npm"]
 USER node
-RUN ["npm", "ci", "--no-optional"]
+RUN ["npm", "ci", "--omit", "optional"]
 COPY --chown=node:node . . 
 
 # STAGE 2: Extend the base image as a builder image
 FROM base as builder
-RUN npm run build:server && rm -rf node_modules && npm ci --no-optional --production
+RUN npm run build:server && rm -rf node_modules && npm ci --omit optional --omit dev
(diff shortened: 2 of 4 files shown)


Urjikoo (approved):
I was able to build container locally on M1 machine and launch it with no issue.

natalynyu (approved):
Looks good - It's building successfully for me. Approved!

PR #562, opened 2022-09-08
Add Room Schedule Table with term dropdown

### LINKED ISSUE #551: Generate Schedule from Room Data
When the user selects a room in the Single Room Schedule, request will be sent to the server with the room id, academic year, and semester, and returns data about courses scheduled in that room.  

When the data is returned from the API, render out the calendar with Course Numbers as the heading of the session block, plus the full course name, faculty names, day of the week, and time range inside the block.

natalynyu:
This PR generates the Room Schedule table. When the Room Schedule page initially loads, the term defaults to the current term. When users interact with the term dropdown and navigate back and forth between pages, the term will remain the same, since the term is saved within the session. There is no default room when the page initially loads, so the user will see a note beneath the term dropdown to select a room in order to see any schedule data. As the user interacts with the room combobox and selects a room, they should begin to see schedule data appear in the table as long as schedule data for that term exists. The look of the schedule is similar to the Course Schedule table. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #551


src/client/api/courses.ts
```diff
@@ -55,6 +56,21 @@ export const getCourseScheduleForSemester = async (
...
+ * information for the courses that will occur in the requested room and
+ * semester.
+ */
+export const getRoomScheduleForSemester = async (
+  roomId: string,
+  calendarYear: number,
+  term: TERM
+): Promise<RoomScheduleResponseDTO[]> => {
+  const response = await request
+    .get(`/api/course-instances/room-schedule?roomId=${roomId}&term=${term}&year=${calendarYear}`);
```
jonseitz, line 70: Instead of using a template string, you can pass the query parameters in the `params` field of a [RequestConfig][1] object as the second argument to `get()`, i.e.

```ts
const response = awaite request
  .get(
    '/api/course-instances/room-schedule',
    {
      params: {
        roomId,
        term,
        year: calendarYear,
      }
    }
  )
```

Functionally it should be exactly the same, but the config object version would, in theory, better handle `undefined` values and uri encoding for special characters. Neither of those is really a concern with how this is currently used, but would be good to avoid in general

[1]: https://axios-http.com/docs/req_config

src/client/components/pages/RoomSchedule/RoomSchedule.tsx
```diff
@@ -35,6 +65,61 @@ const RoomSchedule: FunctionComponent = () => {
...
 
+  /**
+   * Provides the current Academic Year from the server
+   * Later, we may add the current Term to metadata
+   */
+  const { currentAcademicYear, semesters } = useContext(MetadataContext);
+  // This hardcoded academic year will be replaced with the above. Since there
+  // is no current data for Fall 2022, we are using 2020 as the year for now.
+
+  // const currentAcademicYear = 2020;
```
jonseitz, line 76: Can this be removed now?

src/server/courseInstance/courseInstance.service.ts
```diff
@@ -322,4 +327,26 @@ export class CourseInstanceService {
...
+        FacultyListingView,
+        'instructors',
+        'instructors."courseInstanceId" = block."courseInstanceId"'
+      )
+      .where('block."roomId" = :roomId', { roomId })
+      .andWhere('block.term = :term', { term })
+      .andWhere('block."calendarYear" = :calendarYear', { calendarYear })
+      .orderBy('weekday', 'ASC')
+      .addOrderBy('"startHour"', 'ASC')
+      .addOrderBy('"startMinute"', 'ASC')
```
jonseitz, line 349: This also needs `.addOrderBy('instructors."instructorOrder"', 'ASC')` (as on line 137) to make sure that the instructors end up being listed in the correct order in the schedule.

tests/integration/server/courseInstance/courseInstance.controller.test.ts
```diff
@@ -1303,4 +1307,118 @@ describe('CourseInstance API', function () {
       });
     });
   });
+
+  describe('GET /room-schedule', function () {
```
jonseitz, line 1311: As far as I can see, we're never actually testing the `CourseInstanceService#getRoomSchedule` method itself, just the controller behavior. This isn't the worst thing, since the method is mainly relying on TypeORM to assemble the final data object, but it might be worth adding some test for the ordering behavior to ensure that the final output will render appropriately in the client. (I gave myself a headache looking back at the tests I wrote for [the other schedule method][1], so I don't think we need to go into _that_ level of detail).

[1]: https://github.com/seas-computing/course-planner/blob/6d237d1cf2ed1b7e9dc5677ad9723080e1866cf6/tests/integration/server/courseInstance/courseInstance.controller.test.ts#L230-L597

jonseitz (changes requested):
A couple of things noted below, including some notes about testing the service. Per the ongoing team discussion about what's actually worth testing, it may make sense to chat briefly about what those tests should be before proceeding with that...

Otherwise, the Room Schedule tab itself is looking great! You really done a nice job wrangling all the data and putting this together.

tests/integration/server/courseInstance/courseInstance.controller.test.ts
natalynyu: Thanks for your guidance on this Jonathan. I added tests to check the ordering of instructors in commit [45f7693](https://github.com/seas-computing/course-planner/pull/562/commits/45f76939c9fa15f62687ec3e435705836302d520).

jonseitz (approved):
Everything looks great, this is ready to go! :rocket:

PR #567, opened 2022-09-27
Fix Modal Header Displays: Terms & Calendar Years not displaying correctly

### LINKED ISSUE #563: Bugfix: Fix Year Displayed on Courses tab Modals
The Fall modals on the Courses tab are displaying the academic year rather than the calendar year in their headers (e.g. For the academic year 2023, it is displaying "Fall 2023" in the modal instead of "Fall 2022"). This causes confusion for the user in adding ambiguity as to which year is being edited.

natalynyu:
This PR fixes the modal headers so that the Fall term modals display the calendar year as opposed to the calendar year. The displayed terms in the modal header were also updated so that they are in title case. The `getInstanceIdentifier` util function was updated so that it accounts for any casing of the term passed in to the function.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #563

### DIFF
--- a/src/client/components/pages/utils/getInstanceIdentifier.ts
+++ b/src/client/components/pages/utils/getInstanceIdentifier.ts
@@ -1,0 +1,36 @@
+import { TERM } from 'common/constants';
+import CourseInstanceResponseDTO from 'common/dto/courses/CourseInstanceResponse';
+
+interface semester {
+  term: TERM,
+  academicYear: string,
+}
+
+/**
+ * Generates the text to identify the course and term/year information for a
+ * specific course instance and semester. This is frequently used as the
+ * text in modal headers.
+ */
+export const getInstanceIdentifier = (
+  courseInstance: CourseInstanceResponseDTO,
+  semester: semester
+): string => {
+  if (courseInstance && semester) {
+    return semester.term === TERM.FALL
+      ? `${
+        courseInstance.catalogNumber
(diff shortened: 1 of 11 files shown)


src/client/components/pages/utils/getInstanceIdentifier.ts
```diff
@@ -1,5 +1,6 @@
 import { TERM } from 'common/constants';
 import CourseInstanceResponseDTO from 'common/dto/courses/CourseInstanceResponse';
+import { toTitleCase } from 'common/utils/util';
 
 interface semester {
```
Urjikoo, line 5: semester is a type,  so 's' would be capital like `interface Semester`.

Urjikoo (changes requested):
Hi Natalyn this all looks great. i only have one suggestion to capitalize a type. Good work!

src/client/components/pages/utils/getInstanceIdentifier.ts
natalynyu: Great catch, Urji! Thanks for pointing this out. I fixed this in commit [4c53263](https://github.com/seas-computing/course-planner/pull/567/commits/4c5326350bdecec24514014571a3c4cd09d911b1)

Urjikoo (approved):

PR #588, opened 2022-12-11
The course prefix filter buttons style implemented

### LINKED ISSUE #576: Clean up spacing of Schedule Filters
The design and spacing of the header for the schedule needs to be cleaned up, as in [the prototype](https://seas-computing.github.io/course-planner_ux/schedule.html).

@Vittorio2015 to [update prototype to make schedule dropdown more prominent)](https://github.com/seas-computing/course-planner/issues/576#issuecomment-1289593645).

Urjikoo:
## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #576

### DIFF
--- a/src/client/components/pages/Schedule/SchedulePage.tsx
+++ b/src/client/components/pages/Schedule/SchedulePage.tsx
@@ -9,11 +9,11 @@ import {
 } from 'mark-one';
 import { AppMessage, MESSAGE_TYPE, MESSAGE_ACTION } from 'client/classes';
 import { MessageContext, MetadataContext } from 'client/context';
-import { VerticalSpace } from 'client/components/layout';
 import { termEnumToTitleCase } from 'common/utils/termHelperFunctions';
 import { toTitleCase } from 'common/utils/util';
 import { ButtonProps } from 'mark-one/lib/Buttons/Button';
 import styled from 'styled-components';
+import StyledVerticalSpace from 'client/components/layout/VerticalSpace';
 import ScheduleView from './ScheduleView';
 import { useStoredState } from '../../../hooks/useStoredState';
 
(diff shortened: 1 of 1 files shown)


src/client/components/pages/Schedule/SchedulePage.tsx
```diff
@@ -216,9 +231,12 @@ const SchedulePage: FunctionComponent = () => {
 
   return (
     <>
-      <VerticalSpace>
+      <div aria-label="course filter buttons">
+        <span id="Course Filter Button">show/hide catalog prefixes</span>
```
jonseitz, line 235: Per the [HTML spec][mdnhtml], the id field shouldn't contain white space. I'd suggest kebab case (`id="course-filter-buttons"`) instead. 

[mdnhtml]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

jonseitz, line 235: This text should be title-cased: `Show/Hide Catalog Prefixes`

src/client/components/pages/Schedule/SchedulePage.tsx
```diff
@@ -216,9 +231,12 @@ const SchedulePage: FunctionComponent = () => {
...
         {prefixes.map((prefixObj) => (
           <PrefixButton
+            isPrefixActive={isPrefixActive}
+            aria-labelledby="Course Filter Button"
```
jonseitz, line 239: This should be updated to match the kebab-cased id suggested above

src/client/components/pages/Schedule/SchedulePage.tsx
```diff
@@ -46,8 +46,23 @@ export interface PrefixState{
  * A button to filter course prefixes and fade sessionblock based
  * on selected value
  */
-const PrefixButton = styled(Button)<ButtonProps & { prefix: string }>`
- background-color: ${({ prefix }) => getCatPrefixColor(prefix)};
+const PrefixButton = styled(Button) <ButtonProps & { prefix: string, isPrefixActive: (prefix: string) => boolean; }>`
+ margin: 8px;
+ color: ${({ isPrefixActive, prefix }) => (isPrefixActive(prefix) ? 'black' : `${getCatPrefixColor(prefix)}`)};
+ border: ${({ isPrefixActive, prefix }) => (isPrefixActive(prefix) ? '' : `1px solid ${getCatPrefixColor(prefix)}`)};
```
jonseitz, line 52: This `border` value is causing a minor bug where the size of the overall box changes when button are clicked. I've played around with a couple of different solutions, and I think we can either:

1. Don't change the `border`, `color`, and `background-color`, but instead change the `opacity` of the button as a whole to match what's in the table.
2. Add the border to all states on the buttons (including the `:hover` and `:active` pseudo-classes

I'm personally leaning toward 1, but I think either approach is fine.

src/client/components/pages/Schedule/SchedulePage.tsx
```diff
@@ -216,9 +231,12 @@ const SchedulePage: FunctionComponent = () => {
 
   return (
     <>
-      <VerticalSpace>
+      <div aria-label="course filter buttons">
+        <span id="Course Filter Button">show/hide catalog prefixes</span>
         {prefixes.map((prefixObj) => (
           <PrefixButton
+            isPrefixActive={isPrefixActive}
```
jonseitz, line 238: Rather than passing `isPrefixActive` as a function, I'd suggest making it a boolean prop and just calling it with `prefix` here as use time. We can also change the name to just `isActive` to avoid confusion with the function defined above.

```tsx
  isActive={isPrefixActive(prefixObj.prefix)}
``` 

I don't think this makes too much of a different technically, but it does make the component definition easier when you don't need as many embedded calls to `isPrefixActive(prefix)`...

jonseitz (changes requested):
This is looking really good. I have just a handful of style notes and a few things related to conventions we've used elsewhere in the project or in Mark One.

jonseitz (approved):
Pending the tests re-running successfully, this should be all set

PR #592, opened 2022-12-22
SameAs Functionality in Course Entities

### LINKED ISSUE #388: Ensuring SameAs courses can have the same rooms booked
We need to ensure that courses that are the same (e.g. CS109a, AC209a, STATS 121a) may have the same room booked.

In order to do this, we will expect one of the courses to be the canonical course, and the other courses will have their "Same As" field set to that course.

Whenever a query is made to get course instance objects, we will first check the "Same As" field. If this is present, then we will get course instances for that course instead. Otherwise, we will get course instances for the original course.

If changes are made to the course instance on a "Same As" course, these will be reflected on any other related "Same As" courses.

We will not allow setting the "Same As"  field to point to a course that already has a "Same As" defined. This is to prevent circular dependencies.

To do:
* Change "Same As" field in the "Courses admin" to be a autocomplete drop down select box, displaying the course prefix and number.
* This will establish a foreign key relationship between the orginal course and the "same as" course.
* Validate that a course set as "Same as" does not already have a "Same As" relationship set.

Questions for Patrick and Sarah:

* Is this an appropriate workflow?
* Would they want anything on these "Same as" courses to be different between the courses? E.g. Instructor order etc.

jonseitz:
This should the final blocking feature for deployment of the Course Planner. It does require a potentially destructive database schema change, so I'm marking it as breaking -- after merge we can also increment our minor version number.

This replaces the previous free text `sameAs` field in our Course schema with a self-referential foreign key that associates any course with any other "parent" course. It also modifies the existing views and queries to:

1. Substitute the parent course's instances for its own in the Courses Page
2. List all related courses (children, parents, and siblings) in the "Same As" field on the courses page.

I did tweak the put and post endpoints to treat the `sameAs` field as the UUID for the parent course when creating or updating a course, in order to resolve TypeScript conflicts. However, this change has not been reflected in the UI yet, so there is no user-facing way to set the sameAs course for the time being. For that reason I've held off on testing those endpoints until we have a full workflow in place.

There is related work on the [ETL repo](https://github.com/seas-computing/course-planner-etl/tree/feature/586-sameas-courses) that will migrate the relationships from the old course-planner system. Running that will require pulling down the entities from this branch, which can be done by changing the following in `package.json`:

``` diff
diff --git a/package.json b/package.json
index 6786852..dd21684 100644
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "prestart": "./get-entities.sh",
+    "prestart": "./get-entities.sh -b feature/388-sameas-functionality",
     "postinstall": "./get-entities.sh",
     "start": "ts-node -r dotenv/config -r tsconfig-paths/register src/index.ts",
     "lint": "eslint --ext=ts ./src",
```

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High 

## Related Issues:

Fixes #388


src/server/courseInstance/courseInstance.service.ts
```diff
@@ -135,7 +135,13 @@ export class CourseInstanceService {
...
       .addOrderBy('course."numberAlphabetical"', 'ASC')
       .addOrderBy('fall_instructors."instructorOrder"', 'ASC')
-      .addOrderBy('spring_instructors."instructorOrder"', 'ASC');
+      .addOrderBy('spring_instructors."instructorOrder"', 'ASC')
+      .addOrderBy('fall_meetings.day', 'ASC')
+      .addOrderBy('fall_meetings.startTime', 'ASC')
+      .addOrderBy('fall_meetings.endTime', 'ASC')
+      .addOrderBy('spring_meetings.day', 'ASC')
+      .addOrderBy('spring_meetings.startTime', 'ASC')
+      .addOrderBy('spring_meetings.endTime', 'ASC');
```
rmainseas, line 144: Did you meant to commit this as part of the "sameAs" functionality?

rmainseas (changes requested):
This looks great! Just a quick note regarding the `ORDER BY` clause on the course instance service.

src/server/courseInstance/courseInstance.service.ts
jonseitz: Yep, that was intentional. It's not _directly_ related, but it corrects a minor bug with "sameAs" courses showing meetings in a different order. Not the end of the world, but I had [tests that were failing][fail] when comparing the the instances of parents and children (e.g. CS 109A listing Wednesday then Monday but AC 209A listing Monday then Wednesday).

[fail]: https://github.com/seas-computing/course-planner/actions/runs/3743925308/jobs/6356676628

rmainseas: Ah gotcha, okay. I wondered if it was something like that but GH didn't really show me a lot of the code around the change (or even the commit that made the change) so it was a little hard to piece together the context behind it.

PR #604, opened 2023-01-03
Update Placeholder Text for Courses Table

### LINKED ISSUE #599: Courses Page: Text Filter Placeholder Consistency
Two issues:

1. The placeholders for the instructor columns should read "Filter by fall instructors" and "Filter by spring instructors"

2. The placeholder for the Course columns should read "Filter by catalog number".

This might break some tests of the filtering, in which case we should change tests to not rely on specific placeholder text -- Should search by label, preferably, or by column header instead.

natalynyu:
Originally, the placeholder text was created by using the `field` value, which tends to be in camel case. Instead, we would like the placeholder text to be in title case. Using the helper function, `camelCaseToTitleCase`, we can manipulate the `field` and `subfield`s to have the placeholder text show as expected. The reason why we are using the `subField` value is for properties nested under the `fall` and `spring` semesters. For now, we only have one text filter under the `fall` and `spring` semesters (`instructor` column), but we may want to add more text filters in the future. By incorporating the `subField` into the placeholder text, there is a differentiation between different properties nested under `fall` and `spring`.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #599

### DIFF
--- a/src/client/components/pages/Courses/tableFields.tsx
+++ b/src/client/components/pages/Courses/tableFields.tsx
@@ -36,6 +36,7 @@ import { dayEnumToString } from 'common/constants/day';
 import { offeredEnumToString } from 'common/constants/offered';
 import { TermKey } from 'common/constants/term';
 import styled from 'styled-components';
+import { camelCaseToTitleCase } from 'common/utils/util';
 import { PGTime } from '../../../../common/utils/PGTime';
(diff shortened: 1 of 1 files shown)


rmainseas (approved):
Yep. This looks great. Merge away.

PR #606, opened 2023-01-04
Hotfix: Use correct fetch-depth parameter

jonseitz:
The `fetch-depth: 0` option is required to check out the entire git history for the project, so that we can then checkout the specific version we want to build. However, I used `fetch_depth` (with an underscore), which is not a supported property. The action then default to `fetch-depth: 1`, which doesn't include the tag we want to checkout, so the run failed.

In the future we may want to change it to just checkout the specific version from the beginning, but in the interest of creating the smallest change possible we can just fix this line for now.

Unfortunately, this change will only take effect if the workflow is committed to `main`, so I'm opening this as a hotfix against `main`, which we can then pull back into `develop`

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High 

## Related Issues:

Fixes #___

### DIFF
--- a/.github/workflows/05-publish.yml
+++ b/.github/workflows/05-publish.yml
@@ -88,7 +88,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          fetch_depth: 0
+          fetch-depth: 0
       - uses: actions/setup-node@v2
         with:
           node-version: 14


natalynyu (approved):
Looks good to me! Approved.

PR #621, opened 2023-01-11
Fix Course Ordering in Multi Year Plan and Course Admin tables

### LINKED ISSUE #616: Multi-Year/Course Admin: Courses should be sorted numerically
Currently, the Multi-Year Plan and the Course Admin pages sort courses alphabetically by catalog number, which means that a course like CS 50 is appearing after all the 100, 200, and 300 level courses. We should instead be sorting numerically, as we do on the Courses Table. 

There are `numberInteger` and `numberAlphabetical` fields on the `Course` entity that we can use to accomplish this. That should be relatively straightforward to add to the Course admin, but the multi-year plan is using a View that does not include those fields, so I suggest we either:

 - Add `numberInteger` and `numberAlphabetical` to the View (which will need a new migration)
 - Join the `Course` repository to the `MultiYearPlanView` in the service

Then replace 
```
  .addOrderBy('catalogNumber')
```
with 
```
  .addOrderBy('numberInteger')
  .addOrderBy('numberAlphabetical')
```

in `CourseInstanceService#getMultiYearPlan`.

natalynyu:
This PR updates the ordering of courses so that instead of ordering by `catalogNumber`, we are ordering by the integer portion of the course number and the alpha part of the course number so that CS 50 appears before CS 100, for instance. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #616

### DIFF
--- a/src/server/course/course.service.ts
+++ b/src/server/course/course.service.ts
@@ -56,7 +56,8 @@ export class CourseService {
   /**
    * Retrieve all courses in the database and sort by:
    * - area ASC
-   * - catalogNumber ASC
+   * - numberInteger ASC
+   * - numberAlphabetical ASC
    */
   public async findCourses(): Promise<ManageCourseResponseDTO[]> {
     // Any changes to this query should be reflected in FindCoursesQueryResult.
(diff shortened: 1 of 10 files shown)


src/server/course/course.service.ts
```diff
@@ -75,10 +76,13 @@ export class CourseService {
       .addSelect('c."sameAsId"', 'sameAsId')
       .addSelect("CONCAT_WS(' ', p.prefix, p.number)", 'sameAs')
       .addSelect('c.private', 'private')
+      .addSelect('c."numberInteger"', 'numberInteger')
+      .addSelect('c."numberAlphabetical"', 'numberAlphabetical')
       .leftJoinAndSelect(Area, 'a', 'c."areaId" = a.id')
       .leftJoinAndSelect(Course, 'p', 'c."sameAsId" = p.id')
       .orderBy('a.name', 'ASC')
-      .addOrderBy('"catalogNumber"', 'ASC')
+      .addOrderBy('"numberInteger"', 'ASC')
```
jonseitz, line 84: This is good, but I think we need `.addOrderBy('prefix')` before the `numberInteger` to address areas that have courses under multiple prefixes:

jonseitz (changes requested):
This looks great! Just one small change to add ordering by prefix to the Courses query.

src/server/course/course.service.ts
natalynyu: Ah, ok, hadn't thought about that, thanks! I updated the ordering in commit [d69155b](https://github.com/seas-computing/course-planner/pull/621/commits/d69155b337fb61a5cc5d37335cbe559bec40b830) to include ordering by the prefix.

jonseitz (approved):
Looks great! Thanks for getting this sorted out! :rofl:

PR #623, opened 2023-01-11
Create absence records on creation of faculty member

### LINKED ISSUE #620: Faculty: Ensure absences are created along with new Faculty Members
Following on from #619, new absences should be created for all semesters when a new faculty member is created via the Faculty Admin page. In the case of Robert Haussman, it appears that the new absences were not created (or not associated with his entity). We need to investigate and: 

1. Figure out why the absences weren't created.
2. Create the proper absences for Robert Haussman (manually if necessary)

rmainseas:
Creating a new faculty member does not currently populate the absence table. This in turn leads to a white screen of death when trying to access the faculty page.

This PR addresses this issue by adding a [Subscriber](https://orkhan.gitbook.io/typeorm/docs/listeners-and-subscribers) that listens for the creation of `Faculty` and adds the appropriate `Absence` records to match.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #620


src/server/faculty/faculty.controller.ts
```diff
@@ -205,7 +209,11 @@ export class FacultyController {
       jointWith: facultyDto.jointWith,
       notes: facultyDto.notes,
     });
-    const faculty = await this.facultyRepository.save(facultyToCreate);
+
+    const allSemesters = await this.semesterRepository.find();
+    const facultyAbsences = allSemesters
+      .map((semester) => this.absenceRepository.create({ faculty, semester }));
```
natalynyu, line 215: Tagging along our discussion from yesterday's standup, do we need to check that the correct number of absences exist (0) before creating all of them? I suppose since it is a new faculty member, no absences should exist, but I'm wondering about the situation that Jonathan brought up about there possibly being some but not all of the absences already existing and not creating duplicate absence records.

natalynyu (changes requested):
This is looking good, Robert! Do you think it makes sense to add a test to make sure that we are creating the absences as expected?

src/server/faculty/faculty.controller.ts
rmainseas: In this instance, I think that's unlikely as the context behind that initial discussion was the running of the ETL and/or population service where absence records would first by created by the subscriber and then duplicated by the ETL/population service.

In this instance, calling `FacultyController.create` is somewhat of an atomic action in that the faculty member and absence records are created all in one method call.  That is to say that calling `FacultyController.create()` will _always_ result in the creation of a new faculty member. From a data integrity perspective, a new faculty member _always_ gets a new set of absence records.

The only time there's _any_ possibility of this happening at all would be if the faculty creation somehow failed before the absences were created. However, if that were to happen that would cause the absences to created with no `faculty` relation (which is required) which would then in turn cause _that_ database call to fail, thus blocking the creation of absences.

natalynyu: Ok, that makes a lot of sense. I was confused about how we could possibly how some of the absences already existing during our discussion yesterday.

Do you think it would be valuable for us to add a test that makes sure that the right number of absences were created?

rmainseas: In this instance, I would say probably not(at least not in this ticket). I can't immediately think of a situation where this would arise and were that to happen what we might do.

natalynyu (approved):
Looks great, approved!

rmainseas:
Yay, thanks!

PR #626, opened 2023-01-13
Bugfix/619 faculty page rendering fails on null absences

### LINKED ISSUE #619: Faculty Page: Rendering fails on null absences
It appears that the entity faculty member Robert Haussman does not have any associated absences, which is breaking the rendering of the Faculty page. Haussman appears to be a newly added Faculty member, as that name doesn't appear in the old Mongo database.

There are multiple issues here, but in the immediate term we need to update the client code to provide a reasonably fallback if a faculty member doesn't have an associated absence for a semester.

Urjikoo:
## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #619


natalynyu (approved):
Looks great! Approved.

PR #631, opened 2023-01-26
Add version tag via hardcoded .env variable

### LINKED ISSUE #618: Add version tag to front-end code
It would be helpful for our front-end code to include the version of the project at compilation time, so that we can make sure that new deployments have been successful. This value is available in our Github Actions build environment, so we just need to inject it during our webpack compilation stage. I thinking we could either:

1. Add a static `<meta>` tag to our base index.html file at compilation time.
2. Write the version into a js variable, then dynamically insert the meta tag when the page loads.

I'm leaning toward option 2, as that would also give us flexibility if we want to display it in other places (e.g., the page footer, logged to the console, etc.). It's also more indicative of a successful update, since the js code is much more likely to change than the `index.html` file

natalynyu:
This PR adds a `Footer` that contains the app version number via a (currently) hardcoded .env value. The `.env-example` file was updated to include the `APP_VERSION` variable. In the future, the version number will come from Github Actions as opposed to a hardcoded value. 

image

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #618

### DIFF
--- a/src/client/components/App.tsx
+++ b/src/client/components/App.tsx
@@ -120,7 +120,7 @@ const App: FunctionComponent = (): ReactElement => {
           setDataFetching(false);
         });
     }
-  }, [dispatchMessageRef, setDataFetching, appVersion]);
+  }, [dispatchMessageRef, setDataFetching]);
 
   return (
     <div className="app-content">
--- a/src/client/components/App.tsx
+++ b/src/client/components/App.tsx
@@ -34,6 +34,8 @@ import {
   AppHeader,
 } from './layout';
 
+const appVersion = process.env.APP_VERSION;
+
 /**
  * The primary app component. Fetches the current user from the server when it
  * mounts, then saves it to the UserContext to pass down to other components
@@ -76,11 +78,6 @@ const App: FunctionComponent = (): ReactElement => {
   const [isDataFetching, setDataFetching] = useState(true);
 
   /**
-   * Tracks the current version of the application to be displayed in the footer
-   */
-  const [appVersion, setAppVersion] = useState('');
-
-  /**
    * Get the currently authenticated user from the server on launch.
    * If it fails, display a message for the user
    */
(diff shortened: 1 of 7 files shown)


src/client/components/App.tsx
```diff
@@ -70,6 +75,11 @@ const App: FunctionComponent = (): ReactElement => {
    */
   const [isDataFetching, setDataFetching] = useState(true);
 
+  /**
+   * Tracks the current version of the application to be displayed in the footer
+   */
+  const [appVersion, setAppVersion] = useState('');
```
jonseitz, line 81: Since the APP_VERSION variable is never _really_ going to change, I don't think we actually need `useState` here. I think it might make more sense to just declare it outside the scope of the Component (i.e. above line 42) as

```ts
const APP_VERSION = process.env.APP_VERSION;
```
(Note that we can't destructure the value from `process.env`, because we're relying on Webpack to do a simple string replacement; in the compiled code this would be replaced with `const APP_VERSION = "v0.14.1";`)


It's not necessarily a bad thing to make it a state value, but I think doing so implies that there are circumstances where it might change dynamically, when there really aren't.

jonseitz (changes requested):
Just a suggestion about `useState` here, related to the trickiness of the Webpack plugin.

src/client/components/App.tsx
natalynyu: Thanks Jonathan! That's a good point about the version not changing dynamically. I updated the code to declare the application version above the `App` component. I ended up sticking with the variable name `appVersion` so that I did not have to disable the "prefers destructuring" linting rule. (commit [1026ee5](https://github.com/seas-computing/course-planner/pull/631/commits/1026ee5091e293ec9cd8da5d327cc99641a6de5e))

jonseitz (approved):
Looks great! Sorry this got held up by the plague

natalynyu:
> Looks great! Sorry this got held up by the plague

Thanks Jonathan! No worries at all - hope you are feeling better!

PR #632, opened 2023-02-03
Set the app version using the version listed in Github Actions

### LINKED ISSUE #618: Add version tag to front-end code

natalynyu:
In order to get the version that is on Github Actions, we are updating the "publish" workflow to set the value of the APP_VERSION variable to be equal to the value of the Github Actions VERSION_TAG.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #618

### DIFF
--- a/.github/workflows/05-publish.yml
+++ b/.github/workflows/05-publish.yml
@@ -106,6 +106,7 @@ jobs:
         run: |
           rm -rf build
           git checkout ${{ env.VERSION_TAG }}
+          export APP_VERSION="${{ env.VERSION_TAG }}"
           npm run build:client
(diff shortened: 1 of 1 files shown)


rmainseas (approved):
LGTM.

PR #636, opened 2023-02-06
Update course modal to allow same as courses to be set

### LINKED ISSUE #629: Course Admin: Create/Update Modal UI for adding "Same As" field
As a follow-up from #388, we need to handle the new realtionship-style `sameAs` fields in the Course Admin. Currently, I think the Course Admin is going to throw an error if a user tries to add a SameAs relationship on a course, as the server is going to expect a UUID. We need to:

1. Generate a list of valid parent course options with their UUIDs
2. Throw those values in a `Combobox` with a "None" option
3. Allow for changing/removing the parent course
4. Send the UUID field to the server with the other data

We should make sure to apply our validation rules to generate the list of parents, which are:

1. A course can only have one parent
2. The parent course cannot be the child of any other course
3. A child course cannot be the parent of any other course

We should also double-check those rules on the server-side.

rmainseas:
This PR adds front-end UI functionality to set the "Same As" attribute for a given course. This already existed on the server, but was missing the front-end UI component to correctly set that data. This PR does _not_ gracefully disable/hide the modal dropdown to completely prevent the selection of child courses as a parent. This will be provided by another PR/ticket.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #629


src/common/dto/utils/DoesNotMatch.ts
```diff
@@ -0,0 +1,44 @@
...
+      target: object.constructor,
+      propertyName,
+      constraints: [property],
+      options: validationOptions,
+      validator: {
+        validate(value: string, args: ValidationArguments): boolean {
+          const target = args.object as DTO;
+          const [otherProp] = args.constraints as (string & keyof DTO)[];
+          const otherProperty = target[otherProp];
+          if (typeof otherProperty === 'string') {
```
jonseitz, line 33: This makes sense for the particular `sameAs` use case, but for a generic validator we probably want to be able to check for types (e.g. numbers, booleans) right?

If not, we probably want to make it clear that it will only match strings.

src/server/course/course.service.ts
```diff
@@ -73,8 +73,7 @@ export class CourseService {
       .addSelect('c."termPattern"', 'termPattern')
       .addSelect('c."isUndergraduate"', 'isUndergraduate')
       .addSelect('c."isSEAS"', 'isSEAS')
-      .addSelect('c."sameAsId"', 'sameAsId')
-      .addSelect("CONCAT_WS(' ', p.prefix, p.number)", 'sameAs')
```
jonseitz, line 77: Since we're dropping this line we no longer need `"p"` and can also get rid of line `81`, where we `leftJoinAndSelect` on the `sameAsId`

jonseitz (changes requested):
This looks great! Just a couple small notes where things can be cleaned up and streamlined, plus an issue with the list of potential parents being passed in.

src/server/course/course.service.ts
rmainseas: Good point! Fixed in bce2ff7

src/common/dto/utils/DoesNotMatch.ts
rmainseas: Yes, I did wonder about this at the time, but somehow only comparing strings made sense. I think you're right though. Removed that in 1679e590

src/client/components/pages/CourseAdmin.tsx
```diff
@@ -262,6 +262,7 @@ const CourseAdmin: FunctionComponent = function (): ReactElement {
               type: MESSAGE_ACTION.PUSH,
             });
           }}
+          courses={filteredCourses()}
```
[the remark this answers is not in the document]
rmainseas: You're quite right. This is one of those things that made sense in my head when I implemented it along the lines of "We only want to list the final filtered course values in here", but actually after reading your comment, you're entirely correct. That behavior would be frustrating for our users.

Fixed in 5073bf6f

src/client/components/pages/Courses/CourseModal.tsx
```diff
@@ -300,6 +307,19 @@ const CourseModal: FunctionComponent<CourseModalProps> = function ({
       label: area,
     })));
 
+  const courseOptions = [{ value: '', label: '' }]
```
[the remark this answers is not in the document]
rmainseas: Good call. Updated in ee2b0709.

rmainseas:
@jonseitz I think this is ready for re-review. I do want to just call out 045427ab though where I had to convert `value` to `unknown`. I experimented with variations of `typeof`, `Pick` etc. but none of them seemed to do _exactly_ what we wanted so had to resort to casting to `unknown` instead...which in this case I think is probably fine.

jonseitz:
re: the `@DoesNotMatch` decorator and the type of `value`, I think leaving it as `unknown` is fine. 

In looking at this a bit more though I realized that the generic `DTO` type we've been using in our validators is really just being inferred as `{ id: string }`, or "An object with the key you've provided here." So it's not actually checking that the argument is a valid property in the DTO! 

That's not ideal, but I don't think it's actually a huge issue -- if you provided a field that doesn't exist it looks like it would just attempt to compare the decorated property to `undefined`, and end up passing validation (which is technically accurate). We could provide that stricter type-checking by giving an explicit type for `DTO`, e.g. 

```ts
  @DoesNotMatch<UpdateCourseDTO>('id')
  public sameAs?: string;
```

... but that's a little unwieldy and, again, I don't know how valuable it really is. So I'm fine leaving it as it is for now, if you feel the same.

rmainseas:
Yeah, I'm good with that!

jonseitz (approved):
Everything looks good. Approved!

PR #643, opened 2023-02-15
Bugfix/633 remove scrollbar from schedule page

### LINKED ISSUE #633: Scroll bars visible on Schedule page
This is based on feedback from Patrick Ulrich:

    In schedule view, there is now a vertical scroll bar in each box
    when the page loads – which covers the third digit of the course,
    so you can’t actually see any full course numbers.  I don’t recall
    seeing that bar in the demos we did.  Can the boxes be adjusted so
    that the scroll bars are not necessary?  (For example, when you
    click on a course number, then the scroll boxes all go away and
    full numbers are visible across the schedule, as I would expect it
    to be in the default view)

I've gone back to try to get more technical details about browser, OS, etc. and will update this ticket when I've heard more.

Urjikoo:
## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #633

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "tsconfig-paths-webpack-plugin": "^3.2.0",
     "typedoc": "^0.15.6",
     "typescript": "^3.9.6",
-    "webpack": "^4.46.0",
+    "webpack": "^4.35.0",
     "webpack-cli": "^3.3.12",
     "webpack-dev-server": "^3.11.0",
     "webpack-node-externals": "^1.7.2"
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "tsconfig-paths-webpack-plugin": "^3.2.0",
     "typedoc": "^0.15.6",
     "typescript": "^3.9.6",
-    "webpack": "^4.35.0",
+    "webpack": "^4.46.0",
     "webpack-cli": "^3.3.12",
     "webpack-dev-server": "^3.11.0",
     "webpack-node-externals": "^1.7.2"
--- a/src/client/components/pages/Schedule/blocks/SessionBlock.tsx
+++ b/src/client/components/pages/Schedule/blocks/SessionBlock.tsx
@@ -181,7 +181,7 @@ const SessionBlockBodyWrapper = styled.div<SessionBlockBodyWrapperProps>`
       ? 'hidden'
       : 'auto'
   )};
-    +  }
+    }
 `;
 
 /**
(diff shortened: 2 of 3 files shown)


src/client/components/pages/Schedule/blocks/SessionBlock.tsx
```diff
@@ -116,10 +116,14 @@ const SessionBlockWrapper = styled.div<SessionBlockWrapperProps>`
...
-  min-width: 2.5em;
+  min-width: 3.5em;
   position: relative;
   opacity: ${({ isFaded }) => (isFaded ? '0.6' : '1')};
-
+ 
+  ::-webkit-scrollbar {
+    width: 0.2em;
+    height: 0.2em
+}
```
jonseitz, line 126: The `-webkit-scrollbar` pseudo-selector isn't [supported on Firefox][1], so this won't be as helpful to the situation there. I did some searching and found [a different css-only solution on stack overflow][2] that uses the `:hover` pseudo-selector, leaving the `overflow-y` set to hidden when the mouse in elsewhere and `auto` when the mouse is over the block.* That seems to work consistently across Firefox/Edge on Windows and Firefox on Linux. 

In practice, the change looks like:

```diff
 const SessionBlockBodyWrapper = styled.div<SessionBlockBodyWrapperProps>`
-  overflow-y: ${({ isPopoverVisible }): string => (
-    isPopoverVisible
-      ? 'hidden'
-      : 'scroll'
-  )};
+  overflow-y: hidden;
   position: absolute;
   width: 100%;
   top: 2em;
   bottom: 0;
   border-top: 1px solid rgba(255,255,255,0.5);
+  &:hover {
+    overflow-y: ${({ isPopoverVisible }): string => (
+    isPopoverVisible
+      ? 'hidden'
+      : 'auto'
+  )};
+  }
 `;
```


I think it's OK to hold onto the change in `min-width` and the `-webkit-scrollbar` as it does improve the look overall, but I think that change will help us get more consistent behavior across the different browsers/OS combinations.

_* A user can tab through all of the course listings on the page, so it's OK to have this behavior require the mouse._

[1]: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar
[2]: https://stackoverflow.com/a/63348534

package.json
```diff
@@ -137,7 +137,7 @@
     "tsconfig-paths-webpack-plugin": "^3.2.0",
     "typedoc": "^0.15.6",
     "typescript": "^3.9.6",
-    "webpack": "^4.35.0",
+    "webpack": "^4.46.0",
```
jonseitz, line 140: Is there a reason why this needed to be updated as part of this PR? Is there a feature that we needed a new version for? Not a problem to update it, but this is the kind of change would be better done in its own commit, with a message explaining why it was updated, in the (admittedly unlikely) chance that we need to `bisect` and find a breaking change elsewhere in the project.

jonseitz (changes requested):
Left a suggestion about some additional css to make the behavior across systems more consistent, along with a question about the webpage update.

jonseitz (approved):
One small typo that needs to be fixed, but this looks good across all my browser/os testing. 

I'm marking as approved, though that extra `+` does need to be removed before merging.

PR #644, opened 2023-02-16
Roll over no longer active status into new academic year

### LINKED ISSUE #534: Create "No Longer Active" absences in a new academic year
Follow on from #533 

**Adding a new Academic Year**

 - Should create Absence objects for each faculty member
 - If last available semester they were NLA, make them NLA in the new year

rmainwork:
This PR builds on the work in #546 and cascades these changes into the new semester upon creation. This means that:

- A faculty member who is `ABSENCE_TYPE.NO_LONGER_ACTIVE` will cascade into the newly created semester as such
- A faculty member who is anything else will be set as `ABSENCE_TYPE.PRESENT` in the newly created semester.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #534

### DIFF
--- a/tests/integration/server/semester/semester.service.test.ts
+++ b/tests/integration/server/semester/semester.service.test.ts
@@ -19,6 +19,8 @@ import { Semester } from 'server/semester/semester.entity';
 import FakeTimers from '@sinonjs/fake-timers';
 import { MONTH } from 'common/constants/month';
 import { stub } from 'sinon';
+import { Faculty } from 'server/faculty/faculty.entity';
+import { Absence } from 'server/absence/absence.entity';
 import { PopulationModule } from '../../../mocks/database/population/population.module';
 import { TestingStrategy } from '../../../mocks/authentication/testing.strategy';
 import { semesters } from '../../../mocks/database/population/data/semesters';
@@ -26,6 +28,8 @@ import { semesters } from '../../../mocks/database/population/data/semesters';
 describe('Semester Service', function () {
   let testModule: TestingModule;
   let semesterService: SemesterService;
+  let absenceRepository: Repository<Absence>;
+  let facultyRepository: Repository<Faculty>;
   let semesterRepository: Repository<Semester>;
   let fakeConfig: ConfigService;
 
@@ -61,6 +65,8 @@ describe('Semester Service', function () {
 
     semesterService = testModule.get(SemesterService);
     semesterRepository = testModule.get(getRepositoryToken(Semester));
+    absenceRepository = testModule.get(getRepositoryToken(Absence));
+    facultyRepository = testModule.get(getRepositoryToken(Faculty));
     await testModule.createNestApplication()
       .useGlobalPipes(new BadRequestExceptionPipe())
       .init();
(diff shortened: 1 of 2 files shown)


tests/integration/server/semester/semester.service.test.ts
```diff
@@ -377,6 +383,62 @@ describe('Semester Service', function () {
...
+
+          const { absences } = await facultyRepository
+            .createQueryBuilder('f')
+            .leftJoinAndSelect('f.absences', 'absence')
+            .leftJoinAndSelect('absence.semester', 'semester')
+            .where({ id })
+            .andWhere(
+              'semester.academicYear = :acyr',
+              { acyr: testAcademicYear }
+            )
```
jonseitz, line 412: `semester.academicYear` is actually the calendar year while the `addAcademicYear` function is creating Spring in `year` and Fall in `year - 1`. 
So assuming that `testAcademicYear` is 2024, this would create Fall 2023 and Spring 2024, but the `WHERE` clause here would only return the "Spring 2024" semester.

I'd suggest duplicating the test format used above, and query `absenceService` to get the specific spring and fall absences associated with the faculty member. Then you can have separate assertions for each absences that check if `type === ABSENCE_TYPE.NO_LONGER_ACTIVE`

tests/integration/server/semester/semester.service.test.ts
```diff
@@ -377,6 +383,62 @@ describe('Semester Service', function () {
...
+          const { absences } = await facultyRepository
+            .createQueryBuilder('f')
+            .leftJoinAndSelect('f.absences', 'absence')
+            .leftJoinAndSelect('absence.semester', 'semester')
+            .where({ id: presentFacultyMember.id })
+            .andWhere('semester.academicYear = :acyr', { acyr: testAcademicYear })
+            .getOne();
+          const allPresent = absences
+            .every(({ type }) => type === ABSENCE_TYPE.PRESENT);
+          strictEqual(allPresent, true);
```
jonseitz, line 438: Notes on previous test block apply here as well.

jonseitz (changes requested):
There are a couple of issues with the tests, but the functionality here seems to be correct

tests/integration/server/semester/semester.service.test.ts
```diff
@@ -377,6 +383,92 @@ describe('Semester Service', function () {
...
+          strictEqual(newSpringAbsence.type, ABSENCE_TYPE.NO_LONGER_ACTIVE);
+        });
+      });
+      describe(`is anything other than ${ABSENCE_TYPE.NO_LONGER_ACTIVE}`, function () {
+        it(`defaults to ${ABSENCE_TYPE.PRESENT}`, async function () {
+          const presentFacultyMember = await facultyRepository
+            .createQueryBuilder('f')
+            .leftJoinAndSelect('f.absences', 'absence')
+            .where('absence.type = :type', { type: ABSENCE_TYPE.PARENTAL_LEAVE })
+            .getOne();
```
natalynyu, line 439: I think it may be a good idea to rename this variable to `activeFacultyMember` instead of `presentFacultyMember`, since this is supposed to be a faculty member on parental leave.

natalynyu (changes requested):
This all seems to be in working order - I just had a small note about a test variable.

tests/integration/server/semester/semester.service.test.ts
jonseitz: Good point, that lines up better with the test case overall. I made that change in c73bfdd

natalynyu (approved):
Approved!

PR #652, opened 2023-03-14
Incorporate Same As Logic to Update Multi Year Plan and Schedule Table Displays

### LINKED ISSUE #642: Faculty/Plan/Schedule: Apply "sameAs" logic to other interfaces
When a course is marked as a child of another course via the "sameAs" field, we show the course instances from the parent course in place of the child's own instances on the Courses page. This is correct, but that means that faculty and meeting assignments are only being made to the parent course, not to any of the children. This, in turn, means that other pages will only show this information for the parent courses as well:

 - The Schedule will only show parents as taking place on a given day/time 
 - The Faculty page will only show instructors as teaching the parent courses
 - The multi-year plan will only include the instructor names for the parent courses

In all three of those cases (and potentially in others) we want to make sure that we're showing all of the equivalent courses, not just the parents.

In the `CourseInstanceService`, we resolved this by [adding `COALESCE(c."sameAsId", c.id)` to the join clause][coalesce] for the course instances. We would need something roughly similar for the other cases noted above.

Interestingly, this problem is being masked by the way we imported the course data from the old courses planner. We included the instance information for every course BEFORE applying the sameAs relationship, so the existing child instances are still present in the system and are being queried by the current logic. But as the Academic Programs team has been changing parent courses in the system they've noted that their changes are not being carried over to the children. 

There is a related/inverted problem that's specific to those imported courses as well: If a faculty member is removed from the instructor list on a parent courses, the child course won't be updated and will still appear in the Faculty List, Four-Year Plan, etc. Adding the logic above should resolve this issue as well, but we should make sure to check. The specific cases cited by Patrick and Sarah are:

1. Eli Tziperman - was teaching EPS 101 and ESE 101 in Spring 2025; after he was removed from ESE 101 ("parent") the EPS 101 ("child") course is still appearing on the faculty page.
2. Gage Hills was teaching CS 148 ("parent") and CS 248 ("child") in Spring 2025; after removal from CS 148 the CS 248 is still there.
3. The 4-year plan shows ES 121 ("parent") was being offered in the Spring 2025, while AM 121 ("child") is being offered in the Fall 2024. They should both be offered in the spring only.

[coalesce]: https://github.com/seas-computing/course-planner/blob/main/src/server/courseInstance/courseInstance.service.ts#L86

natalynyu:
This PR contains 3 migrations. It updates the server logic such that the Multi Year Plan table will show the instructors of the course's parent if there is one. It also updates the Schedule such that the meeting information (day, time, and room information) of the parent, if there is one, is displayed.

A separate PR will be opened to update the display of the Faculty table to show the same as relationship. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #642

### DIFF
--- a/src/client/components/pages/Schedule/ScheduleView.tsx
+++ b/src/client/components/pages/Schedule/ScheduleView.tsx
@@ -82,6 +82,16 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
    * Track the prefix and catalog number of the course whose details should be shown.
    */
   const [currentPopover, setCurrentPopover] = useState('');
+
+  /**
+   * Keeps track of the course number that was clicked on so that we can use
+   * this value to control the popover visibility
+   */
+  const [clickedCourseInfo, setClickedCourseInfo] = useState({
+    prefix: '',
+    number: '',
+  });
+
   return (
     <WeekBlock
       firstHour={firstHour}
@@ -155,7 +165,9 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                           xOffset="0.5rem"
                           yOffset={`-${2 + listIndex}rem`}
                           title={`${coursePrefix} ${courseNumber}`}
-                          isVisible={currentPopover === instanceId}
+                          isVisible={(currentPopover === instanceId)
+                            && (clickedCourseInfo.number === courseNumber)
+                            && (clickedCourseInfo.prefix === coursePrefix)}
                         >
                           <p>{day}</p>
                           <p>{`${displayStartTime} - ${displayEndTime}`}</p>
@@ -190,7 +202,9 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                     return (
                       <CourseListing key={meetingId}>
                         <CourseListingButton
-                          isHighlighted={popoverInBlock && isSelected}
+                          isHighlighted={(popoverInBlock && isSelected)
+                            && (clickedCourseInfo.number === courseNumber)
+                            && (clickedCourseInfo.prefix === coursePrefix)}
                           disabled={
                             !isSelectedDegreeProgram
                             || (popoverInBlock && !isSelected)
@@ -206,6 +220,10 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                               setCurrentPopover((current) => (
                                 current === instanceId ? null : instanceId
                               ));
+                              setClickedCourseInfo({
+                                prefix: coursePrefix,
+                                number: event.currentTarget.innerText,
+                              });
                             }
                           }}
                         >
--- a/src/client/components/pages/Schedule/ScheduleView.tsx
+++ b/src/client/components/pages/Schedule/ScheduleView.tsx
@@ -83,15 +83,6 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
    */
   const [currentPopover, setCurrentPopover] = useState('');
 
-  /**
-   * Keeps track of the course number that was clicked on so that we can use
-   * this value to control the popover visibility
-   */
-  const [clickedCourseInfo, setClickedCourseInfo] = useState({
-    prefix: '',
-    number: '',
-  });
-
   return (
     <WeekBlock
       firstHour={firstHour}
@@ -165,9 +156,7 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                           xOffset="0.5rem"
                           yOffset={`-${2 + listIndex}rem`}
                           title={`${coursePrefix} ${courseNumber}`}
-                          isVisible={(currentPopover === instanceId)
-                            && (clickedCourseInfo.number === courseNumber)
-                            && (clickedCourseInfo.prefix === coursePrefix)}
+                          isVisible={(currentPopover === instanceId)}
                         >
                           <p>{day}</p>
                           <p>{`${displayStartTime} - ${displayEndTime}`}</p>
@@ -202,9 +191,7 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                     return (
                       <CourseListing key={meetingId}>
                         <CourseListingButton
-                          isHighlighted={(popoverInBlock && isSelected)
-                            && (clickedCourseInfo.number === courseNumber)
-                            && (clickedCourseInfo.prefix === coursePrefix)}
+                          isHighlighted={(popoverInBlock && isSelected)}
                           disabled={
                             !isSelectedDegreeProgram
                             || (popoverInBlock && !isSelected)
@@ -220,10 +207,6 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                               setCurrentPopover((current) => (
                                 current === instanceId ? null : instanceId
                               ));
-                              setClickedCourseInfo({
-                                prefix: coursePrefix,
-                                number: courseNumber,
-                              });
                             }
                           }}
                         >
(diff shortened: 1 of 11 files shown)


tests/integration/e2e/SchedulePage.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+        fireEvent.click(editButton);
+        await renderResult.findByRole('dialog');
+        const sameAsSelect = renderResult.getByLabelText('Same As', { exact: false }) as HTMLSelectElement;
+        fireEvent.change(
+          sameAsSelect,
+          { target: { value: testParentCourse.id } }
+        );
+        const submitButton = renderResult.getByText('Submit');
+        fireEvent.click(submitButton);
+        await wait(() => !renderResult.queryByText('required fields', { exact: false }));
```
jonseitz, line 180: The tests have been failing frequently for me locally. I think it's an issue where the relationship isn't actually being saved before navigating away from the page. I'd suggest using `waitForElementToBeRemoved()` here instead of `wait()`, and to check for the modal by its role instead.

tests/integration/e2e/MultiYearPlan.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+        childRowContent = childRowContent.slice(3);
+
+        // Compare the faculty information and make sure they are equal
+        let areRowsEqual = true;
+        for (let i = 0; i < parentRowContent.length; i++) {
+          if (parentRowContent[i] !== childRowContent[i]) {
+            areRowsEqual = false;
+          }
+        }
+        strictEqual(areRowsEqual, true);
```
jonseitz, line 179: Ideally, we should compare values directly in the assertion, instead of computing and comparing booleans like this. Mainly, this is to give us a higher fidelity error message (i.e. `Expected "A, B, C", got "X, Y, Z"` instead of `Expected true, got false`).

In this case, I think we further want to ensure that the `parentRowContent` and `childRowContent` actually contain data -- otherwise, we could get a passing test with an empty array (or array of empty strings) without actually checking that the child is showing the parent data.

tests/integration/e2e/SchedulePage.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+        await renderResult.findAllByText(parentCatalogNumber, { exact: false });
+
+        // Make sure the child course displays the parent course's meeting information
+        const parentCourseDay = dayEnumToString(testMeeting.day);
+        const parentStartTime = PGTime.toDisplay(testMeeting.startTime);
+        const parentEndTime = PGTime.toDisplay(testMeeting.endTime);
+        const parentLocation = `${testMeeting.room.building.name} ${testMeeting.room.name}`;
+        const parentMeetingInfo = `${parentCourseDay}, ${parentStartTime} to ${parentEndTime} in ${parentLocation}`;
+        const childCourseInfo = `${testChildCourse.prefix} ${testChildCourse.number} on ${parentMeetingInfo}`;
+        await renderResult.findByText(childCourseInfo);
```
jonseitz, line 195: Maybe not the best place to tag this, but I am still seeing the issue where the popover appears for both the child and the parent, with one typically overlapping the other. Looking back through the commit history, I think you were on the right track using a reference to the prefix/number of the clicked course, though we could potentially simplify the setup a bit by just prepending the catalog number to the `instanceId` when we call `setCurrentPopver`, or compare with the state value.

jonseitz (changes requested):
A couple notes on the testing below, including one about the appearance of multiple popovers still being an issue to address.

tests/integration/e2e/MultiYearPlan.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+        fireEvent.click(editButton);
+        await renderResult.findByRole('dialog');
+        const sameAsSelect = renderResult.getByLabelText('Same As', { exact: false }) as HTMLSelectElement;
+        fireEvent.change(
+          sameAsSelect,
+          { target: { value: parentCourse.id } }
+        );
+        const submitButton = renderResult.getByText('Submit');
+        fireEvent.click(submitButton);
+        await wait(() => !renderResult.queryByText('required fields', { exact: false }));
```
jonseitz, line 144: I've also been seeing the same issue here that I noted in the `SchedulePage` test, where the tests are failing because it seems to be navigating before the data is completely saved.

tests/integration/e2e/SchedulePage.test.tsx
natalynyu: Thanks for pointing this out, Jonathan. I updated both the Schedule and Multi Year Plan tests to use `waitForElementToBeRemoved` instead in commit [18c4db7](https://github.com/seas-computing/course-planner/pull/652/commits/18c4db7647edc1d66c9425105001dd4afa479d77).

tests/integration/e2e/MultiYearPlan.test.tsx
natalynyu: This was also updated in commit [18c4db7](https://github.com/seas-computing/course-planner/pull/652/commits/18c4db7647edc1d66c9425105001dd4afa479d77).

tests/integration/e2e/MultiYearPlan.test.tsx
natalynyu: That's a good point about the rows possibly not containing data.

I ended up trying to tackle this from a different direction by finding an instructor first and setting the parent course using the course associated with the instructor I found. I figured this way it would guarantee that the course I was selecting as the parent was guaranteed to be associated with at least one instructor.

I also added a counter variable to check the number of rows that had been compared to further make sure that we were not comparing empty arrays.

For some reason, the multi year plan tests is timing out in the build, but it is passing locally for me over many tries. Do you think there is something that was added that is causing a time out? If not, I could alternatively try separating out the `strictEqual` statements into different tests.

natalynyu: The timeout issue was resolved in [d4f7eed](https://github.com/seas-computing/course-planner/pull/652/commits/d4f7eedb784951766c323b2d3d283ab09dc7a56f) by increasing the timeout limit to 120 seconds.

src/client/components/pages/Schedule/ScheduleView.tsx
```diff
@@ -190,7 +202,9 @@ const ScheduleView: FunctionComponent<ScheduleViewProps> = ({
                     return (
                       <CourseListing key={meetingId}>
                         <CourseListingButton
-                          isHighlighted={popoverInBlock && isSelected}
+                          isHighlighted={(popoverInBlock && isSelected)
```
jonseitz, line 205: I missed this on the first go around, but it looks like there are still some highlighting/fading styles being applied relative to the parent course, e.g. when AC209B is clicked, the other courses in CS109B's block are faded out, but not any of the other courses in other blocks.

jonseitz (changes requested):
The tests are passing for me locally, and it looks like they're passing on CI as well. I suspect that part of the timeout issue has to do with the use of `getAllByRole` --  I remember running into issues with that in the past, having to do with how it parses the DOM tree to find role data...

I did have one last note that I missed on the last review, about how the fading/highlighting properties being applied.

src/client/components/pages/Schedule/ScheduleView.tsx
natalynyu: Thanks for pointing this out! I think that this issue has been fixed via commit [5575662](https://github.com/seas-computing/course-planner/pull/652/commits/5575662160d8d10f4625549b16b0d32ca8cc7610) by adding more conditions determining the values of `popoverInBlock` and `isSelected`.

jonseitz: Great! Thanks for adding the extra context around the change in the commit message (and also for cleaning up that old comment!).

jonseitz (approved):
Everything looks great! Thanks for tackling this.

PR #659, opened 2023-03-20
Show "Check Availability" for all non-seas rooms

### LINKED ISSUE #596: Meeting Modal: Non-SEAS room availability
Currently, Non-SEAS rooms show as available in the Rooms List, but we cannot actually guarantee that those rooms are available. We should instead show "Check Availability." This is currently hard-coded to show when Campus === 'FAS', but we should change "FAS" to "Non-SEAS."

We're no longer using "FAS" as the descriptor for rooms not owned by SEAS, as this could expand the number of campuses greatly if rooms in e.g. HBS or Longwood are added in the future.

This may require changing some tests.

rmainseas:
Any rooms not managed by SEAS were assumed to be FAS. However, given the possible usage of HBS, Longwood etc. rooms in the future....adding a special case for each campus would quickly become unmanagable. Add to that the fact that HBS is _technically_ in Allston.

Therefore, any non-seas rooms are now part of a ficticious "Non-SEAS" campus in the campus table rather than creating special cases for each Harvard school. This still allows the room to be booked, but under the proviso of "We don't have any availability for this room, so you should double check with room book"

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #596

### DIFF
--- a/src/client/components/pages/Courses/RoomSelectionTable.tsx
+++ b/src/client/components/pages/Courses/RoomSelectionTable.tsx
@@ -65,8 +65,8 @@ const displayAvailability = (
   if (meetingTitles.length > 0) {
     return `No (${meetingTitles.join(', ')})`;
   }
-  if (campus === 'FAS') {
-    return 'Check FAS Availability';
+  if (campus === 'Non-SEAS') {
+    return 'Check Availability';
   }
   return 'Yes';
 };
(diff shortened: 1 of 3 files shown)


jonseitz (approved):
Looks good!

PR #662, opened 2023-03-28
Update display of related courses in Faculty Schedule

### LINKED ISSUE #642: Faculty/Plan/Schedule: Apply "sameAs" logic to other interfaces

natalynyu:
This PR updates the display of the Faculty Schedule to include the courses that are related via the "same as" relationship set in the `Course` entity. This is considered a breaking change, because it includes a migration to add the `sameAs` column to the `ScheduleBlockView` so that we can get access to this relationship.  

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #642

### DIFF
--- a/src/client/components/pages/Faculty/FacultyScheduleTable.tsx
+++ b/src/client/components/pages/Faculty/FacultyScheduleTable.tsx
@@ -29,7 +29,6 @@ import {
 } from '@fortawesome/free-solid-svg-icons';
 import {
   FacultyAbsence,
-  FacultyCourse,
   FacultyResponseDTO,
 } from 'common/dto/faculty/FacultyResponse.dto';
 import {
@@ -45,6 +44,7 @@ import {
 import { CellLayout } from 'client/components/general';
 import { MetadataContext } from 'client/context/MetadataContext';
 import { absenceToVariant } from '../utils/absenceToVariant';
+import { deduplicateCourses } from '../utils/deduplicateCourses';
 
 /**
  * Describes the semester specific filter(s)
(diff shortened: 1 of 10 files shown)


Urjikoo (approved):
Looks good!

PR #664, opened 2023-04-04
Feature/656 course report add same as field to spreadsheet

### LINKED ISSUE #656: Course Report: Add SameAs field to spreadsheet
Per discussion with users, they'd like to have a quick way to see which courses have been marked as "SameAs" to others in the spreadsheet. We should duplicate the data that we're currently displaying in the Courses table, possibly with **bolding** or [brackets] around the "primary" course when it's in the list.

Urjikoo:
## Describe your changes
 This PR adds a field to the course report download spread sheet. When a user clicks on `Download Course Report` button, whether the sameAs selected in the `customized view` menu or not, sameAs field appears next to notes column with sameAs information.  

## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #656

### DIFF
--- a/src/server/report/report.service.ts
+++ b/src/server/report/report.service.ts
@@ -51,6 +51,7 @@ export class ReportService {
       { header: 'Is Seas?', key: 'isSEAS' },
       { header: 'Is Undergrad?', key: 'isUndergrad' },
       { header: 'Notes', key: 'notes' },
+      { header: 'SameAs', key: 'sameAs' },
     ];
     yearRange.forEach((year) => {
       const springYear = year.toString().substr(2);
@@ -82,6 +83,7 @@ export class ReportService {
         isSEAS: isSEASEnumToString(course.isSEAS),
         isUndergrad: course.isUndergraduate ? 'Undergraduate' : 'Graduate',
         notes: course.notes,
+        sameAs: course.sameAs,
       };
       yearRange.forEach((year, yearIndex) => {
         const courseInYear = allCourseData[yearIndex][courseIndex];
@@ -128,7 +130,6 @@ export class ReportService {
     (_, index) => startYear + index);
     // Create an object with year keys pointing to lists of faculty info
     const yearToFaculty = await this.facultyService.getAllByYear(yearList);
-
     const facultyToInfoMap: { [id: string]: {
       [year: string]: FacultyResponseDTO } } = {};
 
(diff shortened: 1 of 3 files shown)


src/server/report/report.service.ts
```diff
@@ -82,6 +83,7 @@ export class ReportService {
         isSEAS: isSEASEnumToString(course.isSEAS),
         isUndergrad: course.isUndergraduate ? 'Undergraduate' : 'Graduate',
         notes: course.notes,
+        sameAs: course.sameAs,
```
natalynyu, line 86: The course's `sameAs` field value refers to a course's parent (if one exists). If you take a look at the `Course` entity, the `sameAs` field points to one course. From the description of the ticket, I think that we want to also pull in the course's siblings to the excel report if any exist. The logic to get this information is a bit more complicated. 

I think one way to accomplish this would be to create a separate helper function that returns an array of objects. Each key in the object would be a parent course, and the value would be an array of its the parent course's children. From there you could parse this result to get all of the `sameAs` course information. 

There are likely other ways to tackle this problem, but that was just the first idea that came to mind.

natalynyu (changes requested):
Hi Urji, you're on the right track with this, but we need some additional information to complete the goal of the ticket. The report should list not only the course parent but also any course siblings if they exist. Also, if the course itself is the parent, the `sameAs` excel field should list the course child/children.

src/server/report/report.service.ts
natalynyu: Sorry for the confusion - you can disregard the above comment. The sameAs courses are showing as expected.

src/server/report/report.service.ts
```diff
@@ -51,6 +51,7 @@ export class ReportService {
       { header: 'Is Seas?', key: 'isSEAS' },
       { header: 'Is Undergrad?', key: 'isUndergrad' },
       { header: 'Notes', key: 'notes' },
+      { header: 'SameAs', key: 'sameAs' },
```
natalynyu: The header should probably be "Same As" for the excel sheet.

src/server/report/report.service.ts
natalynyu, line 86: The ticket suggests possibly bolding or putting brackets around the parent course. Was this discussed further?

Urjikoo: Per our discussion, leaving the font style  as is for now.

natalynyu (approved):
LGTM! Merge away when ready.

PR #666, opened 2023-04-06
Fix bug to allow users to unpair sameAs value in Course Modal

### LINKED ISSUE #650: "Same As" Courses cannot be unpaired
It seems that a "sameAs" course cannot be unpaired after being set. When Sarah Colgan tried to remove ES 237 from EPS 237, she was able to save the course but afterwards the two are still linked in the database. This might be an over-eager validation step that's interpreting a blank "SameAs" field as `undefined`, and thus skipping it when updating a course. 

Sarah's email:

> I’m hoping you can help with a specific case: I’d like to unpair ES 237 from EPS 237. I tried to remove ES 237 from EPS 237, but it doesn’t stick.
> These courses were co-listed through spring 2019, but we no longer allow co-listing grad courses across EPS and ES, so I’d like to unpair and then retire ES 237. Going forward, it will only be offered as an EPS course (which will be offered in Spring 2024).

natalynyu:
This fixes a bug in the Course modal that was preventing users from being able to save the `saveAs` value as `null` when the course previously had a non `null` value.

## Types of changes

      to change)

## Checklist:

## Priority:
- [x] Normal 
- [ ] High 

## Related Issues:

Fixes #650

### DIFF
--- a/tests/integration/e2e/SchedulePage.test.tsx
+++ b/tests/integration/e2e/SchedulePage.test.tsx
@@ -193,7 +193,7 @@ describe('End-to-end Schedule Page tests', function () {
         const parentMeetingInfo = `${parentCourseDay}, ${parentStartTime} to ${parentEndTime} in ${parentLocation}`;
         const childCourseInfo = `${testChildCourse.prefix} ${testChildCourse.number} on ${parentMeetingInfo}`;
         await renderResult.findByText(childCourseInfo);
-      }).timeout(120000);
+      });
     });
   });
 });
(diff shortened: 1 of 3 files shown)


tests/integration/e2e/SchedulePage.test.tsx
```diff
@@ -193,7 +193,7 @@ describe('End-to-end Schedule Page tests', function () {
         const parentMeetingInfo = `${parentCourseDay}, ${parentStartTime} to ${parentEndTime} in ${parentLocation}`;
         const childCourseInfo = `${testChildCourse.prefix} ${testChildCourse.number} on ${parentMeetingInfo}`;
         await renderResult.findByText(childCourseInfo);
-      });
+      }).timeout(120000);
```
rmainwork: Unless I'm missing something, I don't think you need this here as [this change is already on `develop`](https://github.com/seas-computing/course-planner/blob/develop/tests/integration/e2e/SchedulePage.test.tsx#L196).

rmainwork (changes requested):
Looks good. Just one change RE: e2e tests.

tests/integration/e2e/SchedulePage.test.tsx
natalynyu: Thanks Robert! This was my attempt at fixing the build issues I was seeing.

rmainwork (changes requested):
This looks great, merge away

rmainwork (approved):

PR #667, opened 2023-04-10
Bugfix/move styled footer component to resolve build issue

Urjikoo:
Initially the new styled footer component `CustomFooter` was initialized in app component because of that the build on github was failing. Therefore it's now initialized right before app component. 

## Types of changes

      to change)

## Checklist:

## Priority:
- [X] Normal 
- [ ] High 

## Related Issues:

Fixes #___

### DIFF
--- a/src/client/components/App.tsx
+++ b/src/client/components/App.tsx
@@ -22,37 +22,20 @@ import {
   MarkOneWrapper,
   PageBody,
   LoadSpinner,
-  Footer,
   ExternalLink,
 } from 'mark-one';
 import { getCurrentUser } from 'client/api';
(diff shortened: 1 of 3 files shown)


natalynyu (changes requested):
As Jonathan suggested in the sprint planning meeting, we should probably move this `CustomFooter` component away from the `App` file. I think it would make sense to create a new file in the `src/client/components/layout` directory.
image

Just as a note, you will need to export in the `index.ts` file like we've done for the other components in the directory:
image

natalynyu (approved):
Looks great! Approved.

PR #669, opened 2023-04-19
Improve course instance offered tests

### LINKED ISSUE #628: Investigate phantom test failures
We continue to see frequent failing test runs, which usually resolve after re-running the test suite a few times. It's almost always a single end-to-end test, though it's not consistently the same one. I thought that #521 and #522 would have fixed it, but the issue still seems to be there. As a course for investigation, I'd like to:

1. Look into most recent unexplainable test failure, and try to triangulate which tests are failing.
2. Look into those specific failing tests, as well as the ones around them in case of aysnc leakages.
3. Watch test runs over the next week to check for continued failures.
4. For future, look into how we might increase performance on test runs, to close up our iteration cycles.

natalynyu:
The original tests related to updating the `offered` field value were not explicitly defining the original offered value, potentially causing typeorm to find retired courses and trying to unretire them. This could be what is potentially causing the tests to fail in the build. These changes explicitly define the original and updated retired values so that the original course instance selected is not one that unexpectedly has a retired value for the offered field.

## Types of changes

      to change)

## Checklist:

## Priority:
- [ ] Normal 
- [x] High 

## Related Issues:

Fixes #628

### DIFF
--- a/tests/integration/e2e/CoursesPage.test.tsx
+++ b/tests/integration/e2e/CoursesPage.test.tsx
@@ -1308,19 +1308,24 @@ describe('End-to-end Course Instance updating', function () {
       let editSpringOfferedButton: HTMLElement;
       let editFallOfferedButton: HTMLElement;
       beforeEach(async function () {
-        const [prefix, number] = courseNumber.split(' ');
-        await courseRepository.findOneOrFail(
-          {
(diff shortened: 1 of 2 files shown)


rmainwork (approved):
This looks great. Merge away!

